PR #24, opened 2026-08-21
feat(storage)!: shard global content for parallel GC

forhappy:
## Summary

- replace flat global xorb and shard keys with canonical two-hex fanout: `.crab/{xorbs,shards}/{first-two-hex}/{full-hash}`
- fail closed on flat, malformed, uppercase, or partition/hash-mismatched global objects
- add provider-aware adaptive bucket-GC listing with three local operator profiles: `adaptive`, `cost`, and `latency`
- keep small/local/serial listings on one recursive stream; fan out only after recursive pagination reaches the complete 256-partition request cost
- expose the policy at initial setup through `crab init --gc-list-profile` and `crab configure --gc-list-profile`, with `crab gc --list-profile` as a one-run override
- preserve the local profile across re-init and reject malformed policy instead of silently overwriting or ignoring it
- move push dedup prefix scans, compact, restripe, metadb, protected push, scoped views, cache routing, and maintenance consumers to the canonical layout
- bump the cache route contract to `crab-cache-service.routes.v2` and document Object Storage Layout V2

## Breaking cutover

This intentionally has no V1 compatibility reader, dual writer, fallback, or migration shim. Existing flat xorb and shard objects must be re-uploaded with the V2 client after old writers are stopped.

## Adaptive cost/latency model

`adaptive` is the default. It starts one recursive stream per global kind. Local stores and effective concurrency 1 remain recursive. Cloud namespaces switch to populated two-hex partitions only after the bounded probe reaches:

- S3/GCS: 256,000 objects (1,000 objects/page × 256 partitions)
- Azure: 1,280,000 objects (5,000 objects/page × 256 partitions)

At the crossover, a full partition scan costs approximately the same number of provider LIST calls as the recursive scan; replaying the probe bounds transition overhead near 2×. Larger namespaces amortize the probe and gain bounded parallel wall time. `cost` always uses one recursive stream per kind. `latency` immediately discovers and scans populated partitions concurrently. Xorb and shard scans share the configured LIST-concurrency semaphore and honor cancellation throughout permit acquisition and stream consumption.

The reported `list_requests` value is explicitly a logical-stream count. Provider pagination and retries remain internal to `object_store` and can issue additional physical API requests.

## Real RustFS qualification

- installed the release CLI using `make install` with an external Cargo target
- initialized and pushed a repository containing a deterministic 5 MiB file
- verified all produced xorbs and shards used matching two-hex partitions and no flat global keys existed
- cloned to a fresh worktree, hydrated all files, and compared BLAKE3 hashes byte-identically
- ran `crab fsck --json`: passed with zero errors
- ran bucket-scope GC dry-run: complete enumeration with zero failures
- deleted the disposable bucket and local fixture and stopped RustFS

Repository policy forbids destructive bucket-scope GC qualification; destructive safety remains covered by in-memory bucket GC tests.

## Adaptive setup smoke

A newly built Crab binary was run in a disposable mounted-volume Git repository:

- `crab init ... --gc-list-profile cost --json` persisted `gc.list_profile = "cost"`
- re-running `crab init --json` preserved `cost`
- `crab configure ... --gc-list-profile latency --dry-run` reported the latency plan
- `crab gc --help` exposed `--list-profile`

## Verification

- rebased `cargo check -p crab --locked`: passed
- bucket GC module: 30 passed, including a 256,001 uniformly distributed object crossover test
- config parse and init persistence/fail-closed regressions: passed
- `cargo test -p crab --locked --lib`: 3,584 passed, 16 unrelated failures, 2 ignored
- `npm run check:links`: 190 pages and 1,960 fragments passed
- `git diff --check`

The broad failures are existing workflow HTTP/checkpoint environment assumptions, missing snapshot baselines, and a replication inventory expectation. No baseline, snapshot, inventory, or workflow expectation was changed to mask them. Rust 1.97 clippy also exposes pre-existing warnings outside this PR's changed surface; CI remains authoritative under the repository toolchain.

## CI status

All GitHub Actions jobs for `c821c69b` failed before executing any step (empty step lists, no job logs, 1–3 second duration). This is runner/infrastructure failure; local rebased compile and scoped tests above are green.

### DIFF
--- a/crab-web/content/docs/cli/cache-service/architecture.mdx
+++ b/crab-web/content/docs/cli/cache-service/architecture.mdx
@@ -29,8 +29,8 @@ Before uploading new chunks, Crab can also ask the cache service which chunks ar
 
 | Object key pattern | Cached |
 |--------------------|--------|
-| `.crab/xorbs/{hash}` | Yes |
-| `.crab/shards/{hash}` | Yes |
+| `.crab/xorbs/{first-two-hex}/{hash}` | Yes |
+| `.crab/shards/{first-two-hex}/{hash}` | Yes |
 | `{repo}/packs/pack-{id}.pack` | Yes |
 | `{repo}/packs/pack-{id}.idx` | Yes |
 | `{repo}/file_index_db/compacted/*.sst` | Yes |
--- a/crab-web/content/docs/cli/reference/crab-configure.mdx
+++ b/crab-web/content/docs/cli/reference/crab-configure.mdx
@@ -55,6 +55,16 @@ Provider-prefixed URLs let Crab infer the provider, so `--provider` is optional
 in those examples. Use it with a canonical `crab://` URL when the provider is
 not S3.
 
+Choose the local bucket-GC listing policy during setup when desired:
+
+```bash
+crab configure s3://team-data/models --gc-list-profile adaptive
+```
+
+The default `adaptive` profile minimizes LIST streams for smaller namespaces
+and switches larger namespaces to hash-partition parallelism. `cost` always
+prefers fewer LIST streams; `latency` immediately prefers partition scans.
+
 ## Tracking options
 
 Pass explicit patterns instead of relying only on the repository scan:
--- a/crab-web/content/docs/cli/reference/crab-gc.mdx
+++ b/crab-web/content/docs/cli/reference/crab-gc.mdx
@@ -32,7 +32,11 @@ can run concurrently with history pruning or history restoration.
 
 Bucket GC reads current and retained historical shard roots for registered
 repositories in parallel, downloads referenced shards with bounded concurrency,
-and lists the complete global shard and xorb namespaces concurrently. It also
+and adaptively lists the complete global shard and xorb namespaces. Small
+namespaces use one recursive stream per kind; larger namespaces switch to
+concurrent populated hash partitions after a bounded provider-aware probe. The
+crossover waits until recursive pagination reaches the cost of the complete
+256-way fan-out, bounding restart overhead to about 2x at the threshold. It also
 tombstones generation-pinned `file_index_db` rows that are outside each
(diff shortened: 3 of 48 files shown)


forhappy:
Kubernetes real-repository qualification (RustFS, isolated bucket, current branch `c821c69b`):

- Source snapshot: `kubernetes/kubernetes@b3bc2ac58fa173967f27ade80f28cc5015b8c1c3`, 31,300 tracked files. Used a self-contained current-tree root commit because only 8.1 GiB remained on the mounted qualification volume; this is not a full 1.8M-object-history benchmark.
- Tracked 11 real Kubernetes artifacts (31.09 MiB total) through Crab; initial push completed in 5.82s. Six independently pushed refs populated 6 xorb partitions and 6 shard partitions.
- Layout audit: 6 xorb + 6 shard keys; every key matched `.crab/{xorbs,shards}/{first-two-hex}/{full-64-hex-hash}`; zero malformed, flat, or prefix/hash mismatches.
- Bucket GC dry-run profile evidence: `adaptive` = 2 logical list requests / parallelism 2 / 0.07s; `cost` = 2 / 2 / 0.07s; `latency` = 14 (2 discovery + 12 populated-prefix streams) / parallelism 12 / 0.07s. All loaded 6 referenced shards and 6 referenced xorbs, with zero deletion candidates.
- Clone completed lazy in 10.57s for 31,301 files; `crab hydrate --all` restored 11 files / 31.09 MiB in 4.79s. All 11 BLAKE3 digests matched the source. `crab fsck --json` passed with 0 errors in 0.10s.
- Separate pre-existing clone/setup behavior surfaced: clone auto-added broad `*.go`, `*.json`, `*.pb`, and `*.png` rules to tracked `.gitattributes`, leaving this fixture dirty. This is outside the two-hex GC/listing change and should be handled separately.

PR #34, opened 2026-08-21
fix: publish concurrent push visibility evidence

forhappy:
## Summary

- publish immutable, content-addressed per-ref visibility edits before activating each ref-journal transaction
- let one generation owner compact those edits into exact Git protocol-v2 visibility evidence without requiring sibling writers' local pack bodies
- hand a late committed transaction to the active manifest compactor, or wait until that transaction is demonstrably covered
- drain at most five visible transaction waves per compaction ownership period, matching the five push-admission slots
- make current-generation Git locator publication survive lock handoff and avoid duplicate repair work when another owner already published proof
- meter every S3 HTTP attempt made by each RustFS push cohort, grouped by bounded request and storage-layout classes without retaining keys or credentials
- retain the backend CAS version returned by lock acquisition and renewal so normal release needs no redundant object read
- skip unnecessary backend-clock PUT/HEAD probes when reacquiring an explicitly released lock tombstone
- shorten only the locator worker-claim and compacted-state commit cadence, avoiding multi-second L0 backpressure in short-lived writers
- suppress SlateDB’s immediate per-writer background GC sweep and amortize locator cleanup across exact 32-generation coverage bands
- retain metadata-owned visibility publication outcome so the compaction owner does not reread its just-published immutable proof

### Lost manifest compactor handoff

A writer committed its ref-journal active marker, released its ref/admission locks, and tried the derived manifest lock once. If another compactor owned that lock, the writer returned successfully. The active compactor could already have captured its snapshot, so the late transaction remained journal-visible but no owner was guaranteed to publish a fresh compacted manifest and protocol-v2 proof.

The late writer now polls its exact active marker while waiting for the derived lock. It exits immediately when another owner removes/covers that marker; otherwise it takes over after the current owner releases the lock. Cancellation and waiting are bounded to two lock TTLs. Because the ref transaction is already committed, cancellation or derived-state failure is reported as repair work rather than falsely failing the push.

While holding the derived lock, the owner drains up to five transaction waves. Five is not an arbitrary retry count: repository admission permits at most five concurrent expensive pushes, so one owner can absorb the entire admitted cohort. The bound prevents a continuous stream from monopolizing the lock; a transaction left active becomes the next owner through the same handoff protocol.

### Lock request amplification

Object-store locks leave holder-checked released tombstones so a stale owner cannot delete or release a newer lease. Two independent costs were present:

1. Reacquiring any existing lock sampled backend time by PUT+HEAD before checking whether its payload was already explicitly released. Backend time is necessary for a live lease because client clocks can be skewed; it has no bearing on an authoritative released tombstone. Reacquisition now skips the clock probe only for explicit releases and proceeds through the existing CAS update. Live and legacy lease expiry still use backend-authored time.
2. Push-admission release discarded the `UpdateVersion` returned by its successful create/update, then GET the same slot to recover a version before writing its tombstone. Admission tickets and the sibling ref/internal lock type now retain the version returned by acquisition and every renewal. Normal release is one conditional PUT. If the token is genuinely stale, the existing fallback reads the holder and refuses to clear a newer owner.

The five admission slots, rotating scan, capacity invariant, wait policy, and stale-owner safety are unchanged. Focused races prove a stale ticket/lock cannot clear a reacquired slot, and the exact single-push meter shows admission release dropping from 2 GET + 2 PUT to 1 GET + 2 PUT.

### Git locator tail latency

Runtime SlateDB tracing corrected the initial hypothesis. Zero-byte WAL-fence discovery completed in milliseconds. The 9.4-second locator phase was L0 backpressure waiting for a previously submitted four-table compaction: the embedded worker and compacted-state commit paths used multi-second default ticks.

The locator keeps SlateDB fencing and the existing L0 bounds. Only its embedded worker claim and compacted-state commit intervals are reduced to 100 ms. The coordinator retains its normal cadence to avoid extra persistent `.compactions` churn. Clean durability flushes are also skipped without weakening the rows-before-coverage invariant.

### Locator garbage-collection lifecycle

Pinned SlateDB 0.14.1 starts every enabled GC ticker with an immediate first tick. A short-lived locator writer therefore scanned every GC directory before its nominal 60-second interval. The locator now disables per-session background GC and runs one foreground pass only when durable exact coverage crosses a 32-generation band. The pass runs after publishing the current reader checkpoint, closing SlateDB, and pruning prior reader checkpoints; it keeps SlateDB’s five-minute minimum age and excludes WAL fences and clone-detach work. GC is acceleration maintenance, so its failure is logged without turning an already committed push into a false failure.

Thirty-two is cost policy rather than a correctness threshold: it amortizes a directory scan across a normal admitted fan-out while bounding superseded locator generations. The exact-coverage transition, not retry count or wall-clock lifetime, selects one owner.

### Redundant visibility-proof reread

Journal compaction already knows whether complete per-transaction evidence produced and uploaded the generation-bound visibility proof before manifest CAS. That outcome was discarded, so the CLI reread the immutable proof after releasing the compaction lock; evidence-less compaction still needed the read/rebuild fallback. `RefJournalCompaction` now carries the durable publication outcome across the metadata boundary. Evidence-backed owners trust the ordering they just completed, while evidence-less and sibling-owner paths retain the existing verification and repair behavior.

Three exact-source 8-writer cohorts showed the expected attribution: historical runs followed `visibility GETs = 16 + compacted generations`; the new runs were one or two GETs lower, matching the number of successful owners reaching the post-lock path. Totals were 895, 1,120, and 1,068 attempts, with every push, protocol-v2 read, content check, and fsck green.

## Why this ownership boundary

Concurrent push writers own only their local object closure. Immutable per-transaction edits let a single generation owner combine sibling evidence deterministically. Retrying a client-side global rebuild is both more expensive and unable to rely on sibling writers' local objects.

Manifest and locator compaction remain derived state. Ref-journal activation is the push visibility boundary; fresh protocol-v2 proof and exact locator coverage follow under repository-scoped owners with bounded handoff.

## Compatibility

The ref-journal visibility evidence is an unreleased main-only format, so there is no shipped fallback reader. The existing reverse-index reconstruction helper remains intentionally narrow: v1.0.14 and earlier are tagged releases that can contain committed `.idx` objects without `.rev`; rebuilding deterministic `.rev` evidence is their migration path, not a second push implementation.

Changing `renew` to require a mutable lease is an internal workspace API adjustment needed to retain the new backend version. All workspace callers, including the CLI repair path and auth service, now use that single lifecycle.

## Live RustFS verification and cost

Final source was tested against `http://127.0.0.1:9000` with bucket/access/secret `crab`:

- 8 simultaneous independent branch pushes all succeeded
- 8 fresh protocol-v2 clones read byte-identical content and passed strict Git fsck
- 4 divergent `main` pushes produced exactly one winner and three structured non-fast-forward results in the hot-ref qualification
- the winning `main` tip was visible to a fresh protocol-v2 clone
- Crab fsck reported zero errors
- slowest branch push fell from about 11.3 seconds in the traced latency baseline to 2.7 seconds before adding the diagnostic proxy

The meter brackets only push processes. It observes SDK retries and LIST pages at the HTTP boundary; later clone/fsck verification and direct AWS CLI inventory are excluded.

Three equivalent 8-branch runs measured 982–1,222 attempts, or 122.75–152.75 per successful push. The exact final-source run measured:

- 1,133 HTTP attempts total, 141.625 per successful push
- per push: 79.5 GET, 13.625 HEAD, 28.375 ordinary PUT, 13.125 LIST, two multipart initiations, two parts, two completions, and one batch delete
- 22,059 request-body bytes and 88,090 response-body bytes per push for this tiny fixture
- +149 net live objects / +93,042 bytes, or 18.625 objects / 11,630.25 bytes per push
- 92 admission-lock requests, 44 manifest-lock requests, and 12 locator-lock requests for the cohort
- 251 locator-DB requests across compacted state, compactions, GC, manifest, and WAL; this remains the largest bounded optimization target

The spread is real background SlateDB/compaction and contention scheduling, so a single low run is not a stable cost promise. The deterministic admission improvement is isolated by the one-push control: one fewer GET per successful short push, while renewal now keeps that saving for long pushes instead of paying a failed CAS before fallback.

A later exact-source locator-lifecycle run kept all 8 concurrent pushes, 8 fresh protocol-v2 clone/content checks, and fsck green. Its seed push fell from 209 to 167 total attempts and locator-GC reads fell from 40 to 28. The 8-way fan-out used 1,178 attempts (147.25 per push), above the previous 1,133, because this schedule elected more locator/manifest compaction owners. The seed isolates the immediate-GC saving; the fan-out variance shows why the optimization is not a flat per-push cost promise.

For a concrete request-only reference, current [AWS S3 pricing](https://aws.amazon.com/s3/pricing/) charges S3 Standard PUT/COPY/POST/LIST at about $0.005 per 1,000 and GET-class requests at about $0.0004 per 1,000 in common US regions; DELETE is free. Applying those rates to the observed independent-push mixes is approximately **$0.00023–$0.00028 per push** (0.023–0.028 cents), or about **$10–$12 for 44,000 pushes/month**. The 4-way hot-ref cohort remains about **$0.000986 per one successful update**, including its three rejected contenders.

Those figures are request charges, not total cost. Storage and retention, actual Git/xorb payload size, clone/fetch traffic, internet or cross-region egress, replication, provider free tiers, observability, minimum billable object sizes in colder tiers, lifecycle work, backup, and operations/on-call dominate a serious team decision. The harness reports the request and inventory inputs so each deployment can apply its actual provider, region, topology, and workload.

## Verification

- five request-meter and boundary regression tests and `git diff --check`
- `cargo test -p crab-coordination --features object-store-lock` — 75 passed
- `cargo test -p crab-metadata --all-features --locked` — 224 passed
- `cargo check -p crab-metadata --no-default-features --locked`
- `cargo clippy -p crab-metadata --all-features --lib --no-deps --locked -- -D warnings`
- `cargo clippy -p crab-coordination --features object-store-lock --all-targets --no-deps -- -D warnings`
- `cargo check -p crab -p crab-auth-server --features crab/gix-transport`
- `cargo build -p crab --features gix-transport`
- `cargo test -p crab --lib git::push::tests --locked` — 236 passed, 1 environment-gated test ignored
- focused late-compactor handoff, stale-release, renewal, cancellation-after-manifest-CAS, visibility-publication outcome, and request-category regressions — passed
- final exact-source metered 8-branch RustFS run — passed protocol-v2 clone, byte-content, Crab fsck, and strict Git fsck checks
- prior 4-hot-ref RustFS run — exactly one winner, three expected rejections, fresh-clone visibility, and fsck checks passed

Repository-wide strict Clippy is currently blocked by unrelated Rust 1.97 baseline findings (including 330 existing `crab` library lints and existing auth-server test lints). The touched coordination crate is clean; GitHub-hosted checks remain authoritative for broad platform coverage.

## Remaining follow-ups

- reduce the remaining pinned-SlateDB GC-boundary reads and variable manifest/compaction-owner amplification without adding a second metadata path; the tested 25 ms coalescing delay reduced generations but was rejected because it added handoff latency without a stable total-request win
- reduce failed admission scans under sustained fanout only if capacity utilization and fairness can be preserved; probing fewer than all five slots is not an acceptable shortcut
- add sustained soak, mid-upload timeout, marker-write failure, and packet-loss qualification beyond the bounded 8-writer and 4-writer swarms
- qualify released artifacts against AWS/GCS/Azure and publish a provider-specific calculator covering storage, egress, replication, minimum sizes, and retention—not request rates alone

## Post-marker SIGKILL recovery

A process killed immediately after RustFS durably accepted the ref-journal active marker left the new ref readable, but its live 21-second ref lease fenced the next writer until expiry. The baseline needed 16 attempts and 25.397 seconds; the production default would expose the same availability gap for roughly its 300-second lease lifetime.

Each ref-journal edit now binds the exact ref-lock holder. A contender may release that lock only after resolving the head to a visible transaction containing the same ref and holder, followed by a holder-checked lock CAS. Prepared transactions remain invisible, unrelated holders remain fenced, and a CAS race cannot clear a successor lock. This is the final ref-lock-protected boundary: the original writer performs no ref-critical work after marker visibility.

The repeated RustFS run recovered on the first successor attempt in 3.655 seconds, before the 21-second expiry. The full crash phase fell from 663 to 507 S3 HTTP attempts while exact `ls-remote`, protocol-v2 clone, byte content, strict Git fsck, and Crab fsck remained green. Normal pushes add no object-store request; they only add the holder identity to the already-written immutable transaction body. The proof reads and conditional tombstone PUT occur only when acquisition encounters a held ref lock.

Additional proof: four acquisition tests, 16 push-lock tests, nine ref-journal tests, minimal and storage/object-lock feature checks, and a five-test crash-meter harness suite pass. Package-owned strict Clippy passes; repository-wide strict Clippy remains blocked by the unrelated Rust 1.97 baseline described above.

## Pre-marker SIGKILL qualification

The RustFS meter now also blocks the prepared ref-head response after the upstream write succeeds, then SIGKILLs the full Git/helper process group before the active marker can be written. The ref remained absent from `ls-remote`; the first successor returned retryable lock contention with a 20-second hint; after the 21-second lease boundary, the second attempt replaced abandoned prepared state and published the exact tip. Protocol-v2 clone, byte content, strict Git fsck, and Crab fsck stayed green.

Honoring the structured retry hint reduced this phase from 16 retries and 698 S3 attempts to two retries and 512 S3 attempts. The final run recovered the pre-marker push in 23.205 seconds and the post-marker push on its first attempt in 3.792 seconds. This proves the intended asymmetry: unfinished transactions retain fencing until TTL, while visible transactions permit immediate exact-holder recovery.


forhappy:
Latest locator-lifecycle increment: commit `36a634f2` is pushed. Focused proof is green (`crab-metadata` all-features: 222 tests; minimal-feature check; strict library Clippy; Crab build), and exact-source RustFS qualification passed 8/8 pushes, 8/8 protocol-v2 clone/content checks, and fsck. The deterministic seed reduced locator GC reads 40→28 and total attempts 209→167; the 8-way cohort was 1,178 attempts (147.25/push), showing compaction-owner variance remains. The failed **Multi-crate guardrails** job reports architecture-inventory drift on files/manifests unchanged by this PR; the latest five `main` Architecture Gates runs also fail, including run 32534725528. Remaining PR checks are still running.

forhappy:
Pushed `4903f20c` (`fix: retain visibility publication outcome`). Metadata compaction now returns whether it durably uploaded generation-bound Git visibility before manifest CAS; evidence-backed owners skip the redundant post-lock proof GET, while evidence-less/sibling-owner paths retain verification and repair. Proof: 224 metadata tests, 236 push-module tests (1 environment-gated ignore), minimal metadata check, strict metadata-library Clippy, Crab build, and three fresh 8-writer RustFS cohorts all green. Meter attribution: new visibility GET counts were 1–2 below the historical `16 + compacted generations` relationship, exactly matching owners reaching the post-lock path; cohort totals were 895, 1,120, and 1,068 attempts. A separately tested 25 ms coalescing delay was discarded after six cohorts because its generation reduction did not produce a stable total-request win and increased handoff latency.

forhappy:
Crash-boundary qualification update (`ca0433cb`):

- generalized the storage cancellation fixture so tests can stop immediately after either the ref-journal active marker or the derived manifest CAS;
- added a restart regression proving that cancellation at the durable active-marker boundary leaves the ref readable through the journal while protocol v2 remains withheld;
- proved a fresh sibling writer compacts the interrupted transaction, recovers the first writer's exact locator rows from immutable pack evidence, publishes complete visibility, and restores protocol-v2 availability.

Proof: `cargo test -p crab git::push::tests --lib --locked` (237 passed, 1 DynamoDB-local test ignored), exact new regression passed, `cargo fmt --all -- --check`, and `git diff --check`. Strict Clippy remains blocked by existing workspace-wide Rust 1.97 lint debt outside this change (first failures in `crab-vfs` and existing `crab` command modules).

forhappy:
Post-rebase update:

- Added post-marker SIGKILL recovery in `217aaa67` and enabled that RustFS phase in CI in `20eda2ff`.
- Live before/after: 25.397s / 16 attempts / 663 S3 attempts became 3.655s / 1 attempt / 507 S3 attempts, with exact ref visibility, protocol-v2 clone, content, strict Git fsck, and Crab fsck green.
- The current `Multi-crate guardrails` failure is baseline architecture-inventory drift unrelated to this PR: its reported `Cargo.toml` dependency/feature mismatches are unchanged from `origin/main`; this PR changes no Cargo manifest. Package-owned checks for the touched coordination and metadata crates pass.

forhappy:
Added pre-marker SIGKILL qualification in `df5efd24`. The RustFS suite now proves both sides of the boundary: invisible prepared state remains fenced until TTL, while visible committed state reclaims its exact holder immediately. Final measured run: pre-marker 23.205s / 2 attempts / 512 S3 attempts; post-marker 3.792s / 1 attempt / 527 S3 attempts. Both completed protocol-v2 clone, content, strict Git fsck, and Crab fsck checks.

forhappy:
Cost and concurrency qualification update (fd27843f):

- Fixed request-meter attribution for S3 LIST calls: the repository prefix is in the query string, not the URL path. Regression test added; 6 proxy tests pass.
- Four isolated RustFS first-push runs were stable at 165–168 S3 attempts, 23–24 KB request bodies, 55–57 KB response bodies, and 38 net live objects / about 12.4 KB.
- Four concurrent independent branches: all 4 pushed, cloned through protocol v2, and passed strict fsck; 479 attempts total (119.75 per accepted push).
- Four concurrent writers integrating onto one branch: all 4 landed and were visible; 1,666 attempts total (416.5 per accepted push), with 1,027 attempts (61.6%) under git_locator_db.
- At S3 Standard US request rates ($0.005/1,000 write/LIST class and $0.0004/1,000 read class), the measured isolated tiny push is about $0.000406 in request charges; the integrated same-ref cohort is about $0.000710 per accepted update. Transfer, retained storage, free tiers, and region-specific pricing are excluded.

Remaining cost priority is locator session request amplification and long-run object-count/GC steady state. Remaining failure qualification is mid-upload timeout, marker-write failure, and packet-loss/retry chaos.

forhappy:
Hardened active-marker storage failure handling in `a8c14cfd` and `32812c24`.

Live RustFS fault qualification (`codex-marker-faults-03`) now proves:
- repeated pre-upstream active-marker 503: ref remains invisible, per-ref JSON is `transient`/`retryable: true`, and a clean retry succeeds in 924 ms (before the 21 s lock TTL);
- response loss after the immutable marker is durably stored: exact-byte create reconciliation returns success, and protocol-v2 clone reads the exact commit/content;
- the injector only simulates post-commit response loss for upstream 2xx and does not mask a 412 conflict.

CI now runs both marker fault cases alongside the existing pre/post-marker SIGKILL cases. Focused Rust tests, 9 proxy tests, Python compile, formatting, debug build, YAML parse, and two live RustFS runs pass. `cargo clippy -p crab --lib --locked -- -D warnings` is blocked only by five pre-existing Rust 1.97 style lints in `crates/crab-vfs`, which matches `origin/main`.

forhappy:
Push request-cost hardening is now in `57ffc96e`.

Live RustFS evidence (`crab/crab` credentials, fresh prefixes):

- 4 same-branch integrated agents: 4/4 `ok`, protocol-v2 clone and exact content passed, strict fsck passed. Locator request budget passed at **121.25 locator attempts/successful push** against a CI maximum of 180; total S3 attempts were 277.5/push.
- The 500 ms SlateDB compactor cadence reduced the earlier 4-agent locator cohort from 886 to 465 locator attempts (-47.5%) and summed push wall time from 46.9 s to 17.9 s (-61.8%).
- 32 same-branch integrated agents: 32/32 `ok`, all content visible, strict fsck passed, and the generation-32 foreground GC boundary ran (2 GC boundary writes plus 365 batch deletions). Typed transient/throttling retries recovered the prior 22/32 saturation failure.
- The 32-agent cost is still not acceptable as a solved result: 25,211 S3 attempts (787.8/push), including 17,065 locator attempts (533.3/push), 94 command integration retries, and a 592 s slowest push. The design roadmap now explicitly keeps coalesced locator publication / replacement of per-generation SlateDB sessions as the top remaining cost item.

New proof/gates:

- CI same-branch mode now integrates all 4 agents and enforces `--max-locator-requests-per-success 180`.
- `cmd::push::tests`: 41 passed.
- locator-focused `crab-metadata`: 9 passed.
- request-meter Python tests: 11 passed.
- `cargo fmt --all -- --check` and `git diff --check`: passed.

forhappy:
CI note: `Multi-crate guardrails` failed on existing dependency-budget drift (crab-xet, crab-coordination, crab-cache, crab-cache-store, crab-metadata, and crab-auth). This branch changes no Cargo manifests or guardrail inventories, and the latest `origin/main` Architecture Gates run at `97e0480a` also failed: https://github.com/crabbuild/crab-oss/actions/runs/32548109262. The changed workflow syntax check passes.

PR #52, opened 2026-08-22
perf(push): harden concurrent publication and protected receive

forhappy:
## Summary

- add `crab metadb owner` as the single renewable per-repository owner for locator publication and Git visibility coverage
- publish immutable per-pack locator evidence, then compact it only through the generation owner
- move repository admission behind per-ref ownership, so same-ref waiters do not consume shared capacity or repeat scans
- keep five bounded reusable CAS slots for direct pushes; charge xorb worker width plus estimated xorb and Git payload, cap ordinary payloads at four slots so small work can still proceed, and retain bounded `Retry-After` cooldown
- stream `git rev-list` directly into `git cat-file` for constant-memory Git-object estimates compatible with the supported Git 2.27 baseline
- keep protected admission service-owned with authenticated repository identity and workload estimates; session-private staging stores do not create ineffective client-local slots
- remove the protected-finalize duplicate body download; reuse verified source materialization and publish `.idx`/`.rev` evidence from the verified Git workspace
- bind protected receive evidence to plan, source generation/ETag, prepare record, repository, and session; clean it after success or TTL expiry
- treat a valid orphaned v1 ref-only visibility proof as absent so the current digest-bound proof can replace it; protocol v2 enters the bounded repair path while digest-bound corruption still fails closed
- document ownership, request cost, cleanup, admission accounting, the shipped legacy Crab Auth contract, and its team-quota limitation

#51 is merged. This branch is based directly on current `origin/main`.

## Verification

- `cargo test -p crab-auth-server --locked` (74 passed)
- `cargo clippy -p crab-auth-server --lib --locked -- -D warnings`
- `cargo test -p crab-coordination --features object-store-lock --locked` (83 passed)
- `cargo clippy -p crab-coordination --features object-store-lock --all-targets --locked -- -D warnings`
- `cargo check -p crab-coordination --no-default-features --locked`
- `cargo test -p crab-metadata --features storage --locked git_visibility` (13 passed)
- focused direct-push tests cover per-ref admission, weighted capacity, xorb/Git memory charging, post-commit release, same-ref isolation, and throttling cooldown
- focused visibility tests cover orphaned-v1 replacement, matching-v1 backfill, and protocol-v2 bounded repair
- focused Git estimator tests cover exclusion of old reachable history in both direct and protected paths
- filtered protected-receive proof reads only the pack trailer plus verified `.idx`/`.rev`, not the full pack body
- `npm run check:links` (190 pages)
- architecture, protocol, RustFS, cross-platform, and package CI rerunning on commit `b738f3b2`

## Known current-main gates

- strict whole-Crab clippy stops in unchanged `crab-vfs` duration-unit lints and unchanged `crab-metadata` test lints before providing a clean whole-workspace result; touched production crates and focused checks pass
- CI reproduces `active_active_push_commits_refs_then_materializes_manifest` failing on the unchanged current-main fixture because its synthetic SHA has no Git object for the required visibility traversal
- the RustFS cache smoke expects an origin xorb GET, while current main validates a hash-checked cache body with a canonical-origin HEAD; the assertion is stale rather than proof of a missing canonical object

## Admission boundary

Direct object-store admission provides repository workload fairness inside one cloud-credential trust domain; it cannot derive a trustworthy team identity. Managed protected preparation resolves the authenticated logical repository and organization, accepts byte/object estimates, and owns quota/concurrency admission before issuing staging credentials. The shipped Crab Auth compatibility endpoint remains supported for existing self-hosted protected-push deployments, but its older prepare contract has neither a stable organization ID nor a workload plan and therefore does not claim team-quota fairness.


forhappy:
Added `perf(push): admit only ref owners` (`0cfbae56`). Repository-wide upload admission now starts only after the per-ref owner acquires its lease, refreshes base state, and rules out an under-lock no-op. Same-ref waiters therefore no longer scan/reserve the five admission slots.

Additional proof:
- `same_ref_waiter_does_not_touch_repository_admission`: waiter fails at `lock` with zero admission objects
- `push_admission_`: 3 passed
- `push_batch_single_ref_succeeds`: passed
- `push_batch_multiple_refs_all_succeed`: passed
- `cargo check -p crab --locked`: passed

Strict workspace clippy remains blocked by unrelated current-tree lint debt (including `crab-vfs` duration-unit lints); no new warning was found in the admission change. The pre-existing `crab/src/git/push.rs` payload warning remains outside this commit.

forhappy:
Added the final admission hardening in `ff7792a4`: the five reusable CAS objects are now capacity units rather than five equal pushes. Xorb pushes reserve one unit per eight configured upload workers (rounded up, capped at all five); no-xorb pushes reserve one. Partial weighted reservations roll back before retry/error, and mixed-weight concurrency tests prove live reservations never exceed capacity. Proof: all 82 `crab-coordination` object-store-lock tests, strict coordination clippy, four Crab admission tests, and the same-ref/no-admission integration test pass.

PR #88, opened 2026-08-26
fix(lfs): make Crab LFS direct-only

forhappy:
## Summary

Make Crab LFS a direct object-storage capability. Users run `crab lfs` locally; Crab does not ship or require a deployable LFS gateway.

- Remove the `crab-lfs-server` crate, deployment files, workflow, HTTP qualification harness, gateway ADR, and gateway release gates.
- Keep and harden native Crab LFS filters and the Git LFS standalone custom-transfer agent.
- Resolve LFS storage lazily only after a non-lazy smudge misses the local cache, so clean and cache-hit operations do not require remote configuration.
- Accept the Git LFS standalone `remote` field as either a Git remote name or direct URL.
- Report initialization failures through the custom-transfer protocol and configure installation with explicit `concurrent=true` and `direction=both`.
- Keep LFS object storage and native lock operations direct. Standard HTTP Batch/basic/File Locking, SSH, and TUS integrations remain external-server profiles.

## Verification

- `cargo fmt --all -- --check`: passed
- `cargo check --workspace --locked --offline`: passed
- `cargo test -p crab-lfs --locked --offline`: 32 passed
- `cargo test -p crab --lib lfs::transfer_agent --locked --offline`: 19 passed
- `cargo test -p crab --lib git::filter_process --locked --offline`: 35 passed
- `cargo test -p crab --lib cmd::lfs --locked --offline`: 159 passed
- `cargo clippy -p crab-lfs --all-targets --locked --offline -- -D warnings`: passed
- Split-crate interface and behavior checks: passed
- Committed-tree architecture check: passed
- `make install`: passed
- RustFS direct qualification run `lfs-direct-scale-20260826-184400`: Git LFS 3.7.1, 100 commits, 1,000 files/objects, 1–10 MiB each, 5,749,027,765 logical bytes; ref equality, byte identity for all 1,000 objects, source/clone `fsck`, and clean clone/fetch/checkout passed.

## Compatibility boundary

Supported: native Crab LFS and unmodified Git LFS using the standalone custom-transfer agent with direct object-storage access. The standalone protocol follows the Git LFS custom-transfer contract documented at https://github.com/git-lfs/git-lfs/blob/main/docs/custom-transfers.md.

Not claimed: standard HTTP Batch/basic/File Locking, SSH/TUS adapters, or managed short-lived grants. Stock `git lfs lock`, `unlock`, and `locks` require an external Git LFS File Locking HTTP API; Crab does not provide that server.

## Validation notes

The broad default Crab test suite still hits the existing `git::remote_helper::tests::multi_push_batch` stack overflow; the isolated test passes with `RUST_MIN_STACK=8388608`. Full Crab clippy still reports existing `crab-vfs` lint violations outside this change.


forhappy:
Follow-up compatibility pass (ae72fe16):

- Git LFS clean now stages only in the local cache; remote publication remains in pre-push/push, so `git add` does not depend on a remote.
- `crab lfs push --object-id` now reads the cached object size before upload, matching Git LFS behavior.
- Local proof: 3,747 Crab library tests passed; real `git add` filter-process smoke passed; real RustFS upload plus object HEAD verification passed.

Remaining compatibility boundary: standard HTTP Basic Batch/basic transfer, verify, ranges, and locking are qualified. SSH discovery/authentication, generic action-based custom transfers/TUS resumability, distributed admission/quota state, and broader provider/OS/client compatibility remain follow-up work for a universal Git LFS replacement.

PR #104, opened 2026-08-28
fix: harden NFS Xet mounts and operator workflow

forhappy:
## Summary

- allow Git partial-clone promisor fetches to retrieve exact reachable objects while retaining visible-ref closure authorization
- classify lazily fetched Crab pointers before NFS size, read, and promotion decisions
- replace the global overlay mutation mutex with shared local and cross-process publish leases so independent paths write concurrently
- retain same-path and intersecting-subtree serialization in the VFS engine
- make commit, export, and reset take an exclusive lease that drains in-flight mutations before snapshotting the overlay
- record post-start mount publish failures and preserve the overlay for retry
- extend the native macOS RustFS/NFS smoke with deterministic concurrent writer processes and end-to-end hash and pointer verification

## Why this is the best fix

This keeps mutation ordering at the existing VFS ownership boundary and changes only the overlay publish barrier. Independent backing files no longer contend on one global mutex, while publish still gets one stable snapshot across threads, store handles, and processes. It does not add another mount mode, fallback path, compatibility shim, or configuration surface.

The Git hardening restores the canonical promisor-object path instead of adding a clone or hydration fallback. Exact-object requests remain authorized against the generation-pinned visible-ref catalog, and arbitrary unfiltered SHA wants remain denied.

## Exact rebased live proof

Native macOS NFS against isolated RustFS, bucket `crabbuild`:

- remote: `crab://crabbuild/mount-large-macos/nfs-concurrent-20260827-rebased`
- published commit: `f893b0aea0411fc3cab2f517ddcc098fba27c594`
- three identical 256 MiB seed files backed by Xet; 768 MiB logical stored as 268,624,156 xorb bytes
- four independent 64 MiB writer processes overlapped for 1.447 seconds
- concurrent aggregate throughput: 172.67 MiB/s through the native mount
- 320 MiB created file, 48 MiB sparse file, truncate, rename, delete, symlink, chmod, diff, export, and reset
- 10,000 ordinary directory entries enumerated
- forced post-commit push failure retained the commit and overlay; retry pushed and cleared the overlay
- every concurrent file hash matched after export, fresh clone, and cold rehydration
- Git blobs and dehydrated worktree files were verified as Crab pointers before hydration
- full scenario completed in 143.01 seconds
- emitted `macos_mount_large_e2e=ok` and `macos_mount_large_rustfs_smoke=ok`

A separate pre-rebase run also passed with 3.839 seconds of writer overlap and 66.23 MiB/s aggregate throughput.

## Focused validation

- `cargo test -p crab-vfs --features nfs`: 458 passed
- `cargo clippy -p crab-vfs --features nfs --lib -- -D warnings`
- `cargo check -p crab --no-default-features --features nfs --locked`
- `cargo fmt --all --check`
- concurrent writer probe self-test and Python compile check
- macOS RustFS/NFS smoke passed twice, including once after rebasing onto current `origin/main`
- Next.js production build passed
- docs link crawl checked 314 HTML pages and 3,192 fragments
- shell syntax and `git diff --check`

## Remaining Level 5 gates

This now proves concurrent independent-path writers on one native macOS NFS mount and a stable commit/export/reset barrier. Same-file overlapping byte-range writes still require application coordination. Separate mounts remain independent working trees. Linux NFS, real cloud providers, crash and reboot recovery, network partitions, longer soak tests, and security qualification remain separate production gates. The cold hydration run also emitted recoverable file-index repair warnings; correctness passed, but eliminating that repair path remains operational follow-up.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2377,6 +2377,7 @@ dependencies = [
  "crab-types",
  "crab-xet",
  "dashmap",
+ "fs4",
  "fuser",
  "gix-hash",
  "gix-object",
--- a/crab/docs/architecture/virtual-filesystem.md
+++ b/crab/docs/architecture/virtual-filesystem.md
@@ -93,6 +93,12 @@ modifications are stored in an overlay database plus an upper directory and are
 merged with the immutable snapshot view. The overlay tracks creates, writes,
 truncates, mode changes, symlinks, unlinks, and directory renames.
 
+Mutations take a shared, cross-process publish lease. Independent paths can
+write their backing files concurrently, while the VFS engine serializes the
+same path and intersecting ancestor/subtree mutations. Publish, export, and
+reset take the exclusive lease, wait for in-flight backing-file mutations, and
+prevent a partial overlay snapshot from being observed.
+
 Overlay writes remain local until the user runs a publish operation:
 - `crab mount diff` inspects the live overlay.
 - `crab mount export` copies overlay changes to a normal directory.
--- a/crab/docs/guides/mount.md
+++ b/crab/docs/guides/mount.md
@@ -310,6 +310,11 @@ crab mount reset --mountpoint /mnt/models --overlay --yes
 checks that the mounted base ref has not moved, creates a Git commit, refreshes
 the mounted snapshot, and clears the overlay after a successful local commit.
 
+Processes using one mount may write independent paths concurrently. Mutations
+to the same path or intersecting directory subtree are serialized. Commit,
+export, and reset drain in-flight mutations before taking a stable overlay
+snapshot.
+
 For large crab-tracked files, commit streams overlay file content directly into
 Crab staging when `.gitattributes` is stable, then writes pointer files into the
 Git index. If `.gitattributes` changed in the overlay, commit first materializes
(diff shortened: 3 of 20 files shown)


forhappy:
Documentation hardening added in `5494600a`:

- expanded the `crab mount` CLI reference with a correct read-only first mount, backend selection, safe writable workflow, concurrency contract, disk budgeting, local commit versus push, retry recovery, static snapshots, and live JSON health checks
- corrected stale CLI guidance: unsupported positional mount syntax, two simultaneous mounts of one repository cache, local-pointer network assumptions, and FUSE-only lifecycle wording
- added the 3,000-word Library guide **How do you operate a Crab mount safely end to end?** covering preflight, Xet range reads, overlays, independent concurrent writers, publish barriers, commit/push, recovery, refresh/switch/reset, monitoring, and production qualification

Web proof:

- `npm run build`: 235 static pages, including `/library/crab-mount-end-to-end`
- `npm run typecheck`: passed
- `npm run test`: 9 passed
- `npm run lint`: passed with 17 pre-existing warnings and no errors
- `npm run check:links`: 322 HTML pages and 3,272 fragments passed
- `git diff --check`: passed

forhappy:
Added the requested visual learning layer to the Crab mount library entry in `24c072dc`:

- interactive mount lifecycle console for cold read, first write, commit + push, and failure/retry
- interactive writer-ownership lab for independent files, same-file writes, and separate mounts
- mounted read-path cutaway and local disk-capacity workbench placed beside the relevant explanations
- explicit transition into the existing scored knowledge check, which records library completion
- responsive layouts, keyboard focus states, 44px interaction targets, and reduced-motion-safe transitions

Validation:
- `npm run build` — 235 static pages
- `npm run typecheck`
- `npm run test` — 9 tests
- `npm run lint` — 0 errors; 17 pre-existing warnings
- `npm run check:links` — 322 pages / 3272 fragments
- headless Chrome visual inspection of `/library/crab-mount-end-to-end`

PR #96, opened 2026-08-27
perf(git): optimize large-repository maintenance and team fanout

forhappy:
## Summary

This PR hardens Crab's Git remote, derived-index maintenance, and qualification path for Kubernetes-scale repositories and concurrent teams. It applies the same broad shape used by mature large-repository hosts—immutable packs, generation-bound catalogs/commit graphs, bitmap-like reachability, geometric consolidation, generated response-pack caching, and bounded admission—within Crab's object-store-only architecture.

The current branch is rebased on `origin/main`. It contains 80 patch commits across 65 files. The roadmap is the detailed ownership, phase, acceptance, evidence, and remaining-gate record: `plans/001-large-repository-scale-roadmap.md`.

### Bounded derived state and pack growth

- Push acknowledgment publishes immutable packs and bounded metadata; repository-sized catalog, commit-graph, shallow-closure, repack, and stale-membership work belongs to the generation owner.
- The object catalog, visibility proof, shallow closure, and split commit graph are generation-bound and published together. Reader repair remains lazy and uses non-blocking writer probes.
- Automatic consolidation is geometric and suffix-bounded. Stable large packs remain untouched; current serving inventory converges while immutable recovery history remains separately rooted for GC.
- Locator startup, large-batch lookup, stale-pack sweeps, pack sidecars, source acquisition, and selected response assembly all have bounded concurrency/read-ahead/body-size contracts.
- Complete response packs preserve reusable deltas where safe; filtered and shallow responses retain exact object-set and Git fsck verification.

### Fetch and cache path

- Upload-pack uses catalog/visibility/commit-graph acceleration without materializing the repository-wide OID universe on ordinary reads.
- Absolute shallow closures are indexed for depth 1/10/100/1,000. Ordinary fetches from an already-shallow client reuse the exact generation transition when there is no deepen request; explicit deepen, relative deepen, incomplete tag state, and other contextual requests retain the conservative planner.
- Generated response packs are keyed by exact generation, selection, filter, protocol, and authorization identity. Equivalent dense requests canonicalize their OID selection so team fanout coalesces.
- Cross-process cache waiters poll the immutable descriptor with deterministic jitter. Distributed ownership is probed only near the observed owner's expiry or at an approximately one-minute jittered ceiling. Cancellation and expiry takeover remain bounded; initial configuration/authentication errors still fail fast.
[...]

PR #109, opened 2026-08-28
fix: verify and harden Linux NFS Xet parity

forhappy:
## Summary

- make interactive native NFS mounts run the existing cache-aware remote refresh path and avoid publishing duplicate generations when the remote ref is unchanged
- turn the large RustFS/Xet smoke into an explicit native Linux NFS qualification with mount-table proof, read-only/manual refresh, automatic refresh, writable overlay reset/export, four overlapping independent writers, push-failure recovery, commit/push, and fresh eager-clone verification
- fix the Docker worktree revision probe and add retained Linux RustFS/Xet evidence on the NFS workflow

## End-to-end proof

Executed locally in a privileged Linux arm64 Docker container against RustFS with access key/secret `crab` and bucket `crabbuild`:

- native mount confirmed as `127.0.0.1:/crab ... nfs`
- 32 MiB seeded Xet file read byte-for-byte through NFS
- manual read-only refresh and 30-second writable automatic refresh succeeded
- four writers overlapped while each wrote and fsynced an independent 4 MiB file
- writable overlay covered modify/create/delete/rename/symlink/sparse-extension operations
- injected post-commit push failure retained the local commit and dirty overlay; retry pushed successfully
- fresh eager clone reproduced every SHA-256 and retained compact 122-byte Git pointers
- RustFS object count: 181
- scenario time after build/setup: 87.885 seconds

Evidence run: `mount-large-nfs-linux-20260828-proof`

## Validation

- `cargo fmt --all --check`
- `cargo test -p crab-vfs --features nfs` — 458 passed
- `cargo check -p crab --bin crab --no-default-features --features nfs`
- `make nfs-smoke-script-check`
- native Linux NFS smoke and report verification
- full Docker Linux NFS + RustFS/Xet E2E

## Concurrency boundary

This qualifies Crab's same-host loopback NFS contract. Independent paths can be written concurrently. Applications still need to coordinate same-file or intersecting-subtree writes, and separate mounts are independent Git working trees rather than a distributed lock service.

### DIFF
--- a/.github/workflows/nfs-mount.yml
+++ b/.github/workflows/nfs-mount.yml
@@ -10,6 +10,7 @@ on:
       - "crab/Cargo.toml"
       - "crab/Makefile"
       - "crab/scripts/release/dispatch-nfs-release-evidence.sh"
+      - "crab/scripts/docker/run-mount-large-rustfs-smoke.sh"
       - "crab/scripts/docker/run-mount-nfs-linux-smoke.sh"
       - "crab/scripts/nfs-evidence-summary.py"
       - "crab/scripts/nfs-read-path-bench-report.py"
@@ -20,6 +21,7 @@ on:
       - "crab/src/cmd/mount.rs"
       - "crab/src/vfs/**"
       - "crab/tests/prop_coordinator.rs"
+      - "crates/crab-vfs/**"
   pull_request:
     paths:
       - ".github/workflows/nfs-mount.yml"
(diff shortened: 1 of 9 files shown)


forhappy:
CI diagnosis: this is a pre-existing deterministic locator request-budget regression on current `main`, not an NFS change or a random retry flake.

Evidence from retained released-shape RustFS reports:

- `main` after #104 (`fecb914a`): 849.5 catalog requests/success, failed the 500 ceiling.
- this PR (`7aa401ab`): 849.5, identical failure.
- PR #96 completed runs: 398.25, 351.75, and 351.75, all below 500.
- all four same-branch pushes in #109 integrated with zero rebase retries; the excess was SlateDB compacted-table discovery, not push correctness.

The first validated correction is `b8c51985` in #96 (`perf(locator): bound large-repo startup fanout`): complete locally resolvable ref tips bypass unnecessary locator catalog startup, while partial/legacy state retains the exact locator path. Its regression and RustFS evidence are already owned by that PR.

I am not raising/suppressing the budget or duplicating the 303-line locator fix in this NFS PR. Once #96 lands, rebase #109 onto updated `main`; rerunning against the current base would reproduce the same failure.

PR #153, opened 2026-09-06
perf(add): enlarge direct classification batches

forhappy:
## Summary
- raise direct xorb classification batches from 32 MiB to 64 MiB
- keep the byte-bounded remote lookup and promotion path intact
- document the v1 hard-cutover batch target

## Verification
- `cargo test -p crab-staging --locked` (197 passed, 1 ignored)
- RustFS end-to-end evidence from the preceding qualification: 2.0625 GiB in-place edit add 12.55s, push 14.75s; fresh-bucket fsck passed with zero errors and isolated-cache clone/hydrate was byte-identical.

### DIFF
--- a/crab/docs/design/add.md
+++ b/crab/docs/design/add.md
@@ -162,7 +162,7 @@ parallel files and later adds cannot assign the same chunk to different local
 xorbs. Only the claim winner feeds its builder. Other recipe occurrences wait
 for and then lease the winner's placement. The coordination is disk-backed and
 batch-bounded; it does not require a repository-sized in-memory hash map.
-Direct preparation classifies up to 32 MiB per remote-index query. Files still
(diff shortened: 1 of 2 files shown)


forhappy:
Follow-up optimization (f23f69291dd): add-time remote candidate classification now uses a bucket/global-prefix scoped advisory cache. It adds a 65,536-entry in-memory LRU plus a 2,000,000-row SQLite cache with batched 512-key reads, bounded eviction, and cache hit/miss debug metrics. Push still revalidates every placement and origin proof; cache failures/staleness fall back to the existing proof lookup/local authority path.

Validation: cargo fmt and git diff --check pass. A new Cargo compile was not run because the required $HOME/Workspace build volume is unavailable on this host per AGENTS.md.

PR #156, opened 2026-09-06
perf(add): harden cached preparation and paged push dedup

forhappy:
## Summary

- Bound the advisory remote-candidate cache across restarts and concurrent writers; preserve negative expiry and invalidate matching misses only after committed publication.
- Preserve local hits and completed remote-proof pages during bounded lookups. Remove the 4,096-candidate push dedup cliff.
- Preserve literal-path symlink boundaries and consolidate batched validation.
- Isolate each concurrent add lookup's receipt snapshot/result. The pre-fix live probe lost 280 remote chunk hits with four workers; after the fix, one and four workers both retain all 1,057 remote hits and prepare only four boundary chunks.
- Add reproducible fresh-bucket RustFS scale qualification with structured phase telemetry, deferred/concurrent add, cold cross-repository dedup, and byte-checked hydration cycles.

## Verification

- Corrected runtime `6e23f81ee16c`: 4,145 library tests passed serially, three ignored; Crab clippy passed; fresh RustFS lifecycle passed 132 checks.
- Earlier boundary changes: 215 staging tests and 96 integration/property/schema tests passed; strict staging clippy and formatting passed.
- Current PR head `ce3fdecfc03`: all 29 executed CI checks passed, eight policy-skipped. Earlier unrelated browser-tooltip contrast retries and parallel Git/LFS fixture-lock failures are documented; no assertions or baselines were weakened.
- Full scale result: **1,223 checks passed, zero failures**. Fresh bucket, 50 × 2 GiB non-zero files, 500 code files, three committed/pushed versions. All 1,100 SHA-256 comparisons passed across cold hydration and rehydration; both dehydration cycles verified all 50 pointers; strict Git fsck passed. This is 100 GiB logical and 20 GiB initial unique entropy, not 100 GiB unique entropy.
- Cold consumer staged 8,274 chunks (> one lookup page), uploaded one new xorb, and cloned/hydrated byte-identically.
- Code-only branch push in the large repository: 1.001 s, no new xorbs, cloned code bytes identical.
- One-worker versus four-worker proof coverage passed again inside the fresh scale run.

| Scale operation | Elapsed |
|---|---:|
| Initial 100 GiB add | 559.033 s |
| Initial push | 593.262 s |
| Edited-version add, v1 / v2 | 620.299 / 642.509 s |
| Edited-version push, v1 / v2 | 245.025 / 277.197 s |
| Deferred 2 GiB prepare / later Git add | 18.800 / 20.120 s |
[...]

PR #161, opened 2026-09-07
perf(staging): batch push retirement and reduce edited-push latency

forhappy:
## Summary

- Batch staging file retirement into one SQLite transaction and run unowned payload inventory cleanup once per batch. Writer/shared-push publication, rollback and recovery share the canonical path.
- Preserve committed segment counts, prepared-payload leases, whole-batch rollback and commit-before-unlink ordering. No format/version, dependency or durability-setting changes.
- Split push timing into retirement, cache/index and manifest subphases.
- Create fresh smoke-test private cache roots with mode 0700; existing directories and production security checks remain unchanged.

## Measured performance

Fresh local RustFS qualification: 50 × 2-GiB large files, ten shared content families, 500 small code files, three versions. Approximately 20 GiB unique base content, not 100 GiB unique entropy.

| Phase | Before | After |
|---|---:|---:|
| First edited push | 245.025 s | 113.972 s |
| Second edited push | 277.197 s | 108.308 s |
| First post-success cleanup/cache | 108.758 s | 29.793 s |
| Second post-success cleanup/cache | 131.536 s | 30.215 s |

Uploaded bytes are identical across the before/after scale runs: 57,579,824 and 57,285,483 bytes. The old aggregate timer included staging retirement, not just cache work. Same-snapshot SQL replay isolated per-file commits versus batching: 16.705 s → 1.686 s. The harness permission correction contributes separately to end-to-end improvements.

## Verification

- Fresh-bucket 100-GiB RustFS lifecycle: **1,223 checks passed**, including **1,100 SHA-256 comparisons**, commit/push, cold clone/hydration, dehydration/rehydration and final Git fsck.
- Separate fresh-bucket lifecycle canary: **132 checks passed**.
- Crab default-feature library: **4,145 passed, 3 ignored**.
- Staging library: **217 passed, 1 ignored**; new whole-batch failure rollback and shared prepared-payload retention regressions.
- Smoke-harness Python suite: **11 passed**.
- Release CLI build, strict staging all-target clippy, Crab CI lint-category gate, workspace format and diff checks passed. Crab lint retains existing CI-allowed warnings.
- Earlier broad runs exposed unchanged timing-sensitive cache-lock/cancellation tests and an overly long temporary-path setup; final normal-temp run after live I/O completed passed without weakening assertions.
- Disposable qualification repositories/caches and all three task-created RustFS buckets removed; reports/logs retained locally.

Detailed evidence, ownership map, binary provenance and limitations: `crab/docs/benchmarks/add-push-hardening-rustfs.md`, section “Edited-push retirement follow-up”. Runtime source tested is this patch atop merged PR #156; the binary embeds its runtime-identical pre-squash revision, documented with SHA-256.

## Design and remaining gaps

Is this the best fix rather than merely plausible? Batching removes repeated durable commits/global scans at the staging owner boundary, instead of weakening durability or deferring cleanup. It also removes duplicated writer/read-only retirement loops. The tradeoff is a longer individual write transaction and potentially more WAL retained until commit. Ownership publication remains a separate transaction with existing retry/recovery.

Retirement still takes approximately 29 seconds at scale; metadata work remains material. Initial pushes and add wall times were slower on the shared workstation/USB SSD, so this does **not** establish improved add latency, universal low latency, all historical-version reconstruction, or production S3/GCS/Azure performance. All formats remain v1; no compatibility path is added.

### DIFF
--- a/crab/docs/architecture/git-protocol-v2.md
+++ b/crab/docs/architecture/git-protocol-v2.md
@@ -179,6 +179,14 @@ The RustFS concurrency qualification follows each independent-ref and hot-ref
 write swarm with fresh protocol-v2 clones, strict Git fsck, and byte checks so
 ref visibility alone cannot satisfy the gate.
 
+Generation-owner contention does not suppress protocol-v2 capability discovery
+when the compacted manifest has visibility evidence, a bounded pending handoff
+bound to that manifest, or is empty. Discovery does not repair or compact
+metadata. Fetch admission waits for owner work when needed, completes and
+validates pending handoffs, then verifies the advertised refs and immutable pack
+inventory against the admitted snapshot. A pending handoff is not itself object
+authorization; absent evidence still withholds v2.
+
 Upload-pack admission is repository-scoped and distributed: each helper
 process must hold one of the fixed object-store read leases for the duration
 of its session. A rotated, jittered retry probes one slot at a time, leases
--- a/crab/docs/benchmarks/add-push-hardening-rustfs.md
+++ b/crab/docs/benchmarks/add-push-hardening-rustfs.md
@@ -115,3 +115,104 @@ scale/probe data and completed canary data/buckets were removed; reports and
 logs were retained. Qualification covers the exercised local RustFS lifecycle,
 not 100 GiB of unique entropy, all historical large-file versions, all CLI
 commands, or production S3/GCS/Azure behavior.
+
+## Edited-push retirement follow-up
+
+The completed 100-GiB qualification reported edited pushes of 245.025 and
+277.197 seconds for 57,579,824 and 57,285,483 uploaded bytes. The aggregate
+`post_success_cache_warm` phase included staging retirement and took 108.758
+and 131.536 seconds; that label did not isolate cache work.
+
+Additional timers now separate lock release, index preparation, negative
+invalidation, index installation, xorb caching and staging retirement, plus
+manifest preparation. No serialized format, version or durability setting
+changes. A smaller reproduction uses the same 50-file/500-code-file shape,
+256 MiB per large file, ten content families and 1-MiB independent edits.
+
+| Local 12.5-GiB reproduction | Uploaded bytes | Push seconds | Retirement seconds | Manifest preparation seconds |
+|---|---:|---:|---:|---:|
+| Original per-file retirement | 57,853,538 | 38.574 | 21.176 | 10.439 |
+| Batched retirement only | 57,796,320 | 25.174 | 3.676 | 11.013 |
+| Batched retirement and valid private cache | 57,494,049 | 16.764 | 1.621 | 3.728 |
+
+The first comparison isolates batching from the cache fixture correction.
+Replaying the same prepared SQLite snapshot with identical deletion SQL took
+16.705 seconds with per-file commits versus 1.686 seconds with one commit.
+The bundled SQLite source documents the default 1,000-page WAL checkpoint
+threshold after commits; repeated commits amplified local index write costs.
+The production fix also runs unowned payload inventory cleanup once per batch.
+
+The smoke harness previously created its cache root with mode 0755 under umask
+022. Crab correctly rejected that root for private proof-cache access. Fresh
+fixture roots now use 0700; existing roots are not chmodded and the product's
+security checks are unchanged. This qualification error must not be attributed
+solely to product retirement performance. Timings use consecutive edits on a
+shared workstation/USB SSD, not an isolated or universally reproducible bound.
+
+| Changed boundary | Entry/caller → owner → callee | Siblings and proof |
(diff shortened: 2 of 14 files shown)


forhappy:
## Fix for branch-fanout protocol-v2 qualification failure

The failed CI reader had correct content and passed strict Git fsck, but the required protocol-v2 packet trace was absent. The original CI log does not record the owner probe, so the exact scheduling interleaving cannot be reconstructed from that artifact alone.

A controlled local RustFS probe reproduced the same symptom with only the generation-owner lease changed: v2 advertised before holding the lease, absent while held (clone content correct and fsck clean), and advertised again after restoring the lease. Eight- and 32-reader ungated baseline runs passed, confirming why an ordinary rerun alone was insufficient evidence.

Removing the owner gate alone passed the CI-shaped replay but a 32-reader run still failed once. A second deterministic gate then paused a reader after the compacted manifest CAS and before catalog materialization: another helper rejected the durable pending handoff and withheld v2. The matching Rust regression also failed before the second fix.

The complete fix removes the owner-activity gate and its diagnostic journal LIST, and recognizes a bounded pending visibility handoff bound to the compacted manifest. The shared metadata parser validates that binding for both admission discovery and handoff application. The existing published-proof predicate is unchanged: owners must not confuse pending evidence with completed publication and skip materialization. Actual fetch admission retains bounded owner/repair waiting, cancellation, complete base/edit validation, generation-pinned visibility authorization, and exact comparison with the discovered refs and pack inventory. No format, dependency, lock or staging-retirement changes.

Evidence map: `dispatch_capabilities` → `snapshot_available` / `capability_snapshot_is_stable` → manifest, published-proof or pending-handoff reads. `read_catalog_pending` is shared with `apply_catalog_journal_edits`; `compact_ref_journal` persists the handoff before manifest CAS. `serve_admitted` → `open_repository_with_visibility_requirement` remains the repair/authorization owner. Empty repositories, missing evidence, active journals, stale locators and replica-selected stores retain their existing boundaries. Current main has both admission gaps.

Is this the best fix rather than merely plausible? Protocol support should not depend on a one-read advisory owner hint or on whether another reader has finished materializing an already-published handoff. Recognizing the existing durable state fixes both gaps without sleeps, retries in the CI assertion, weaker object checks, or a compatibility fallback. Existing owner-aware fetch admission remains canonical.

Regression: the added active-owner assertion in `upload_pack_repairs_locator_for_current_manifest_generation` failed before the fix. The obsolete owner-suppression expectation in the handoff test now checks v2 remains available while retaining its deferred-locator checks. The missing-visibility tests remain unchanged.

Validation: full default-feature Crab library **4,145 passed, 3 ignored**, using CI's 8-MiB test-thread stack. An initial default-stack replay overflowed; no production or test assertion change was used to bypass it. Release build and live replay results follow below.

Intermediate owner-only build `0fe980c59f24f1727e0c05ff1eaed40831b3b7149547de36e6fe4068dca5834b` passed the held-owner clone, 23-check concurrent CI scenario and isolated 92-check partial-clone suite, but is not sufficient final proof because of the subsequent 32-reader failure. An initial partial-clone attempt shared a bucket with concurrent writes and invalidated its whole-bucket dry-run invariant; the isolated replay passed without changing the assertion. Final validation of both fixes follows below.

Final source: **4,145 Crab library tests passed, 3 ignored**, including both new regression assertions and the unchanged missing-evidence rejection cases. **34 metadata visibility tests passed**, including pending target-generation/pack-binding rejection. The metadata no-default-feature check passed with four existing unused-item warnings in `ref_registry.rs`. One broad replay stalled in the unchanged Unix filter test's worker join; its stack was sampled, only that test process was terminated, and the unchanged full replay passed. All temporary diagnostics were removed.

Final optimized binary SHA-256: `ee3fcc6aac034f26f2a4a22c8ad29095dc777552959572d80d9255650aa95c0c`. The held-owner clone and gated pending-handoff probe both passed. Fresh-bucket CI-shaped concurrency passed **23 checks** (including crash/marker faults and request-cost budget), and the separate fresh-bucket partial-clone suite passed **92 checks**. The first 32-reader metered round passed. A subsequent metered stress round hit the local Python proxy's socket-allocation failure (`Can't assign requested address`, surfaced as S3 502), not a protocol-evidence failure; larger follow-up rounds use the existing direct-RustFS mode without the optional meter. The original metered CI scenario is not bypassed.

Both direct 32-reader follow-up rounds passed: **96 successful branch-reader clones across three completed 32-reader rounds**, with content and strict fsck checks. The final Crab CI lint-category gate and workspace format/diff checks passed (existing CI-allowed warnings retained). All disposable run data/caches and five task-created RustFS buckets were removed and bucket absence verified; reports, logs and clearly separated debug probes remain local. Commit: `edb4c405ff4`. Rebased on current main; hosted Linux CI must still run on this commit.

PR #164, opened 2026-09-08
fix(cli): guide pushes to initialize missing remotes

forhappy:
## Summary

- classify a missing canonical layout as an uninitialized remote instead of corruption
- add `CRAB-E0032` with exact `crab init crab://...` recovery in text and structured output
- withhold `crab init` success until the remote layout and generation-0 manifest exist
- leave occupied or malformed prefixes unchanged and remove destructive reset guidance
- update CLI docs, architecture notes, and RustFS expectations

## UX

Before, `crab push` and `git push` reported that the canonical v1 descriptor was missing and suggested resetting isolated development data.

Now they report the configured remote, provide a copy-paste initialization command, and preserve the original command for retry. If remote initialization fails after local setup, `crab init` reports that partial state instead of emitting false success.

## Validation

- focused `crab-metadata` missing-layout test
- `crab-write` initialization test suite
- focused Crab authorization, error rendering, retry, GC, and occupied-prefix tests
- error-code golden test
- `npm run build`
- `npm run check:links` (398 pages, 4305 fragments)

Not run locally: the full RustFS smoke suite, which requires its dedicated object-store environment.

### DIFF
--- a/crab/docs/architecture/git-integration.md
+++ b/crab/docs/architecture/git-integration.md
@@ -75,7 +75,7 @@ and ETag evidence before Crab claims provider parity.
 
 | Git operation or capability | Direct S3-compatible remote | Automated evidence | RustFS evidence |
 |---|---|---|---|
-| Ref list / `git ls-remote` | Supported; missing layout or manifest fails closed until `crab init` creates generation 0 | `git::remote_helper` manifest and resolved-session tests | missing, valid, and malformed manifest cases |
+| Ref list / `git ls-remote` | Supported; an uninitialized remote reports `CRAB-E0032` with the `crab init <REMOTE>` recovery step, while malformed layout or manifest data fails as corruption | `git::remote_helper` manifest and resolved-session tests | missing, valid, and malformed manifest cases |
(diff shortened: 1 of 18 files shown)


forhappy:
CI follow-up: both failed jobs traced to the same stale sibling assertion in `crab-metadata`. The production change intentionally preserves `StorageError::NotFound` for a missing layout, but `snapshot_requires_layout_but_ignores_equivalent_json_formatting` still expected the retired synthetic `CorruptObject`. Commit `36ad733cae2` updates that contract assertion to verify the exact missing descriptor path. The focused repro and the exact five-crate CI command now pass locally.

forhappy:
CI follow-up: the Rust quality failure was a race in `prepopulated_walk` test setup. Each test acquired the `GIT_DIR` mutex only after `git init` and local identity configuration, so a sibling test could redirect those commands to its repository. Commit `3832369f292` now acquires and clears `GIT_DIR` before fixture creation, then points it at the initialized repository. The three-test target passes with an intentionally invalid ambient `GIT_DIR` and global/system Git config disabled.

PR #175, opened 2026-09-10
fix(s3): harden burst load and large-object clones

forhappy:
## Summary

This pull request delivers the production S3 gateway profile for Crab repositories. It presents logical Crab repositories as S3 buckets so existing S3 clients can use their normal endpoint, credentials, and signed-request workflows.

- Supports SigV4 header and presigned-query requests, legacy SigV2, path-style addressing, optional virtual-hosted addressing, conditional reads and writes, metadata, tags, copy, list, range reads, checksums, and durable multipart operations.
- Publishes Crab writes atomically through immutable metadata, visibility proofs, ref leases, and Git LFS/Xet-aware storage. Existing Crab/Xet pointer objects retain chunk-aware, deduplicated range reconstruction.
- Bounds request admission, scratch reservations, multipart staging, cache placement, maintenance work, manifest replay, request-body idle time, response-stream ownership, and readiness repository fanout. Saturation returns retryable `SlowDown` with `Retry-After`.
- Provides liveness/readiness probes, fixed-cardinality Prometheus metrics, credential-safe dependency logging, graceful draining, and hardened read-only/non-root container, Helm, Compose, and ECS Fargate profiles.
- Adds recovery-safe multipart reconciliation, deterministic completion receipts, orphan cleanup, cache/catalog health checks, and supervised repository-maintenance guidance.

## S3 compatibility profile

The gateway implements the documented frozen S3 compatibility profile. It does not claim every AWS control-plane feature; modeled-but-unsupported fields fail closed with `NotImplemented` instead of being silently ignored.

`PutObjectTagging` and `DeleteObjects` XML bodies require and validate either `Content-MD5` or a modeled `x-amz-checksum-*` value. If both are supplied, both are validated, matching current official SDK request shapes.

Gateway-authored large objects use verified Git LFS publication. Existing Crab/Xet pointer objects use chunk-aware range reconstruction and deduplication. Moving new S3 uploads themselves onto Xet remains a separate storage-format migration that requires pointer, GC, clone/hydration, restart, and SDK compatibility proof.

## Hardening in this head

- `0b27f198f03` refreshes `/readyz` repository views in bounded four-repository batches, reducing large-catalog readiness latency without unbounded provider fanout or changing the ten-second deadline.
- `17790390363` bounds idle waits for every `s3s`-buffered XML route, including multipart completion and rejected XML operations, while preserving streamed upload paths.
- `4fae7a9cba3` canonicalizes physical repository prefixes and rejects provider/bucket/prefix aliases before listeners open.
- `67b56aba596` bounds idle waits for XML mutation request bodies before the `s3s` parser buffers them, covering `DeleteObjects` and `PutObjectTagging` without changing the existing `RequestTimeout` behavior for streamed `PutObject` and `UploadPart` bodies.
- `0ea174f9efa` keeps legacy Git paths outside the S3 key profile out of LIST results and preserves bounded continuation scanning.
- `e8ca3fd7076` makes transport parsing explicit: HTTP/1 headers and buffers have fixed bounds and a live header-read deadline; HTTP/2 streams and header lists are bounded and use keep-alive failure detection.
- `913b5d014a8` bounds live S3 connections to four times the request budget, reserves management capacity for probes, and drains listener accept errors before cleanup.
- `96d045a1064` bounds `ListBuckets` materialization to the requested S3 page plus one lookahead, using the documented 10,000 default.
- `65bf7d19eca` rejects idle `PutObject` and `UploadPart` bodies with `RequestTimeout`, releasing spool and admission resources.
- Earlier commits bound multipart/catalog scans, manifest-delta replay, signed-request metadata, read-view singleflight maps, and complete GET response ownership.

## Verification

- Mounted-volume local proof: `cargo fmt --all -- --check`, strict Clippy, `cargo test -p crab-s3-gateway --locked` (143 library tests and 1 binary boundary test), and `git diff --check`.
- The qualification image installs pinned Boto3 in an isolated environment and runs real signed traffic covering put, head, full GET, range GET, tags, conditional writes/deletes, copy, listing, multipart create/upload/list/complete, and multi-delete. It retains a schema-checked, credential-free report.
- RustFS qualification covers SigV4, presigned requests, tamper rejection, exact ranges, Xet reconstruction, large multipart completion, scratch admission, metrics, graceful shutdown, Compose replacement recovery, and multipart crash/expiry reconciliation.
- The pinned Boto3 qualification completes a deterministic 512 MiB object as eight sequential 64 MiB parts, verifies a full GET, and verifies a range crossing a persisted part boundary; the report includes source/full/range SHA-256 values and timings.

Latest workflow runs for head `0b27f198f03`:

| Workflow | Run | Current state |
| --- | --- | --- |
| [S3 gateway](https://github.com/crabbuild/crab/actions/runs/34608915592) | signed Boto3 and gateway qualification | queued |
| [Large repository RustFS qualification](https://github.com/crabbuild/crab/actions/runs/34608915529) | large-repository and RustFS proof | passed |
| [Provider contract qualification](https://github.com/crabbuild/crab/actions/runs/34608915543) | provider contract matrix | passed |
| [CI](https://github.com/crabbuild/crab/actions/runs/34608915746) | repository-wide checks | pending |

## Deliberate remaining limits

Provider-internal wire-attempt telemetry, live GCS/Azure transport credentials, live Kubernetes/Alertmanager delivery, and a direct Crab SDK committed-write throughput baseline remain separate managed-environment qualification work. Bounded admission and maintenance intentionally expose `SlowDown` during saturation; heavily fragmented repositories still require a supervised `crab metadb owner` or explicit `crab repack` pass.

The production-readiness milestone tracks these managed-environment qualifications separately from completed local and RustFS interoperability evidence. Aggregate CI and product-specific checks are still running for the current head `0b27f198f03`.


forhappy:
CI diagnosis and fix pushed in f7cc30582c5. The failing lifecycle test used a 100 ms owner deadline before its mock-server transport barrier; on a loaded runner, the valid Timeout result could win that barrier select. A deterministic 150 ms pre-barrier delay reproduced the exact failure. The test now gives setup a 30 s bound, pauses Tokio time only after the real HTTP request is pending, advances beyond the owner deadline, and still proves the socket reaches EOF before the timeout result returns. Local proof: the injected-delay regression passed, 100/100 repetitions passed, all 4 lifecycle tests passed, the complete crab-sdk --all-features suite passed, strict Clippy passed, and formatting is clean. The accompanying f9e214dd44a commit adds canonical SigV4 streaming-upload qualification and removes the aws-chunked transport token from stored object Content-Encoding metadata, matching S3 behavior.

forhappy:
Temporary SigV4 session credentials are now supported and qualified in `8f96d089e24`.

Behavior:
- `session_token_file` and RFC 3339 `expires_at` form an atomic optional credential pair.
- Missing, wrong, duplicate, or misplaced tokens fail closed; expired tokens return `ExpiredToken`.
- Header and presigned-query tokens must be bound by the verified SigV4 request. Static credentials reject injected session tokens.
- The packaged-image job exercises an unchanged AWS CLI with header auth and a token-bearing presigned URL, checks wrong/missing-token failures, and scans logs/metrics for all credential material.

Local proof (Cargo artifacts under `/Volumes/Workspace/crabbuild-target/crab-5623-s3-gateway-admission`):
- `cargo test -p crab-s3-gateway --locked`: 117 passed
- strict all-target Clippy, `cargo fmt --all --check`, `git diff --check`: passed
- optimized release build: passed
- actionlint 1.7.11 and YAML parse: passed
- exact release binary + fresh RustFS + unchanged AWS CLI: static PUT, temporary header-signed HEAD, token-bearing presigned GET with byte-identical body, wrong/missing token `InvalidToken`, DELETE and post-delete 404: passed

The isolated gateway and RustFS processes were stopped after qualification; generated proof data was moved to Trash.

forhappy:
Retained S3 gateway evidence is now pushed in `a43fd487217`.

The packaged-image gate now produces a versioned `crab.s3-gateway-evidence` report with exact source/diff identity, container image digest, pinned RustFS and client versions, fixture digest, a fixed 20-check inventory, request/byte/latency measurements, RSS and disk usage, and terminal state. The validator fails stale, dirty, incomplete, skipped, unmeasured, or identity-bearing evidence. Failed jobs still emit an incomplete report when the runner reaches the evidence step; only complete reports pass verification. Successful or failed reports are retained for 90 days.

Local proof:
- 8 evidence producer/verifier unit tests passed.
- End-to-end CLI report production and verification passed against a clean source identity.
- actionlint 1.7.11, YAML parse, and `git diff --check` passed.
- Generated local evidence was moved to Trash after verification.

forhappy:
Retained-evidence qualification is green at `9b764bfaed5`: https://github.com/crabbuild/crab/actions/runs/34549979505

Final packaged-image evidence:
- 21/21 required cells passed; none skipped.
- Clean exact checkout and empty diff digest verified.
- Container image digest, pinned RustFS image, AWS CLI/curl/platform versions, and 2 MiB fixture digest retained.
- 21 HTTP requests and 1,341 backend requests measured; backend bytes read/written were positive; resident working set and writable/cache/scratch usage were recorded; active scratch returned to zero.
- Signed catalog, PUT/GET, exact range GET, streaming SigV4 and tamper atomicity, static and temporary presigned/header auth, invalid-token rejection, pressure handling, metrics/leak checks, graceful shutdown, Compose deployment, and multipart restart recovery all passed.
- The uploaded artifact was downloaded and independently re-verified locally, then moved to Trash. No qualification containers remain.

The two preceding failed runs also uploaded correctly incomplete reports. Their root cause was limited to non-portable RSS collection; protocol traffic was isolated and green before the final Docker working-set measurement fix.

forhappy:
Pushed `f13f7d405b6`: direct Crab/Xet range GETs now stream selected verified chunks through a single bounded backpressure slot, cancel source work when the response is dropped, and avoid response-sized scratch. Full GETs still materialize and verify the whole-file hash before headers. Pre-existing Crab/LFS pointers now use deterministic content-digest ETags, so HEAD/list/conditions/range admission no longer hydrate the payload; gateway-authored MD5 and multipart ETags are unchanged. Copy-source ranges keep scratch because mutation requires a verified source before publication.

Local proof on `/Volumes/Workspace`: `crab-read` 166/166, `crab-s3-gateway` 118/118, `crab-sdk --all-features` green including `open_timeout_drains_pending_transport`, strict Clippy green, architecture guardrails green, all 25 split-crate interface slices green, formatting and diff checks green.

forhappy:
Pushed `77ebc6d2ff4`: the packaged-image gate now publishes a real duplicated 64 MiB Crab/Xet object before gateway startup, then uses signed S3 HEAD/list and a throttled presigned 16 MiB range GET. Qualification fails unless both pointers expose the same BLAKE3-derived ETag, the range matches the independently retained source slice, xorbs were actually published, and metadata/range delivery records zero Xet reconstruction scratch. The retained report schema is now v2 with a 24-check inventory.

Local real-runtime proof: the minimal release publisher staged 128 MiB as 2,094 chunks, pushed one deduplicated 64 MiB xorb, and the current gateway returned a byte-exact 16,777,216-byte range (`sha256=212000b235d12fd7dc553e62d516126a048ee5b1c1b238ac16ac31affc281aba`) with identical pointer ETags and zero peak/written Xet scratch. The isolated bucket was deleted and the temporary fixture directory was moved to Trash after verification.

forhappy:
Follow-up pushed in `a50d3f1e48d`: `LocalCache::prepare()` now initializes the canonical catalog schema before the gateway opens its listeners, removing the first-fill/metrics race that could emit `no such table: reservations` on a fresh cache. The regression failed before the fix and passes after it. Verification: 228 local-cache tests, 174 decoded-range-cache sibling tests, 117 gateway library tests plus the binary boundary test, strict Clippy, and a fresh runtime replay with 50 consecutive healthy catalog scrapes. The temporary replay fixture was moved to Trash; the existing RustFS service was unchanged.

forhappy:
Added production-scale object-list pagination in `d9600da64dc`.

What changed:
- `crab-remote-git` now owns a bounded recursive blob-page API. It binary-seeks raw Git trees from the byte prefix/exclusive continuation, preserves complete-path lexical order, collapses `/` delimiter groups before descent, and stops at `MaxKeys + 1`.
- `crab-s3-gateway` no longer walks and sorts the complete repository for every page. It resolves the full S3 marker into repository-relative bytes and loads gateway attributes in one delta-chain pass for only the emitted blobs.
- Existing V1 delegation, root ref listing, multipart listings, object authorization, and signed-request handling remain on their prior owner paths.

Regression proof:
- A real-Git fixture with 1,024 flat entries makes the old exhaustive traversal exceed a deliberately low 64-entry budget; two late-prefix pages of seven objects each succeed under the same budget.
- A gateway-boundary test exercises three V2 pages containing both objects and `CommonPrefixes`, including `a-1`, `a/`, `a0` ordering, continuation reuse, truncation, key counts, and a stored gateway ETag.
- `cargo test -p crab-remote-git --locked`: 126 unit + 80 real-repository + 1 doctest passed.
- `cargo test -p crab-s3-gateway --locked`: 118 library + 1 binary-boundary test passed.
- Strict Clippy for both crates, formatting, diff checks, and all 25 split-crate interface checks passed.
- A clean external-target rebuild of the gateway pagination regression passed after reclaiming only this worktree's disposable Cargo artifacts.

The exact-SHA S3 qualification is now queued at https://github.com/crabbuild/crab/actions/runs/34558178801. The large-repository RustFS and provider-contract workflows are already green at this SHA.

forhappy:
Updated the PR at `f3107507d6b` with retained large-list qualification. The packaged release image passed [S3 gateway run 34559177260](https://github.com/crabbuild/crab/actions/runs/34559177260): schema v3 verified all 25 checks; signed AWS CLI traversed 10,032 logical 64 MiB objects (627 GiB) in exactly 11 pages in 8.740 seconds, with exact ordering, no duplicates, exact late-prefix continuation, two delimiter groups, one shared Git pointer blob, and zero Xet hydration scratch. Milestone [S3 Gateway Production Readiness](https://github.com/crabbuild/crab/milestone/1) remains assigned.

forhappy:
Follow-up on the transient SDK performance check: [Rust SDK run 34559177232](https://github.com/crabbuild/crab/actions/runs/34559177232) passed on the failed-job rerun. The retained rerun measured SDK/core cold wall ratio 0.871 and warm ratio 0.854, with identical origin request/byte ratios of 1.0 and 438.5 MiB peak RSS. The first attempt had crossed only the cold timing threshold (1.114 vs 1.10); its integrity, origin I/O, warm timing, and memory checks were otherwise sound.

forhappy:
Schema-v5 production-hardening evidence passed on the clean merge checkout.

- S3 qualification: https://github.com/crabbuild/crab/actions/runs/34563395746
- 26/26 retained assertions passed.
- Large listing: 10,032 objects in 11 pages; gateway 9.103 s versus direct RustFS SDK 11.373 s (80.1%); ordered and unique; zero metadata scratch.
- Multipart registration: 8 MiB and 64 MiB parts each retained an 871-byte durable state record (0-byte delta, below the 64 KiB bound); both staged payloads were reclaimed after abort.
- Runtime: 116,811,366-byte resident working set, zero active scratch bytes, and zero writable-container-layer bytes.

forhappy:
Schema-v6 production-hardening evidence is green in [S3 gateway run 34568524532](https://github.com/crabbuild/crab/actions/runs/34568524532): 27/27 retained assertions on a clean merge checkout. The forced-kill gate exited the packaged gateway with code 137, reclaimed exactly one orphan in the first completed cleanup scan, preserved the live Open and frozen Completing uploads with one part each, removed the orphan state and capacity reservation, and reclaimed all fixture payloads through signed S3 cleanup. The same run listed 10,032 objects through 11 bounded pages in 8.984s versus 11.854s through direct RustFS (75.8%), retained 871-byte multipart state for both 8 MiB and 64 MiB parts, used 5,586,812 bytes RSS, and left zero active scratch and zero container-writable bytes.

forhappy:
Deployment-hardening update in `d1880b8ece3`: added `crates/crab-s3-gateway/deploy/ecs/crab-s3-gateway.yaml` and parameter examples for an immutable-image, two-task ECS Fargate service. The template creates ECR/logging/IAM/ALB/service resources while requiring caller-owned VPC, subnets, security groups, ACM certificate, backend bucket/prefix, and Secrets Manager values. It uses a read-only non-root task, separate scratch/cache volumes, least-privilege prefix-scoped S3 access, TLS termination, readiness health checks, 100% minimum healthy deployment, and rollback circuit breaking. Local `cfn-lint 1.41.0`, YAML/JSON, actionlint, and contract checks pass; live Fargate qualification remains intentionally unclaimed.

forhappy:
The ECS validator is now green in the current S3 gateway workflow: `Validate ECS Fargate template` passed with pinned `cfn-lint==1.41.0`, YAML contract checks, resource-inventory checks, digest/non-root/read-only checks, readiness-port 8081 verification, Fargate CPU/memory rule presence, and placeholder/example coverage. The packaged gateway qualification remains in progress.

PR #192, opened 2026-09-12
feat(http-server): harden multi-cloud deployments

forhappy:
## Summary

This makes `crab-http-server` deployable as one provider-neutral service while keeping production claims gated by real evidence.

### Documentation and onboarding

- rewrite `REFERENCE.md` and the former `WRITE-DESIGN.md` into structured, lengthy references with diagrams, tables, examples, and navigable contracts; rename the design document to `DESIGN.md`
- document two supported starts: one-command Docker Compose for local evaluation, and Terraform-generated provider values plus one Helm chart for teams
- replace per-repository server configuration with a durable CAS catalog and `repository create`, `adopt`, `list`, and `set-members`; every replica discovers catalog changes without restart

### Portable team deployment

- add one strict Helm chart for EKS, GKE, and AKS with typed generated configuration, immutable images, two-replica minimum, hard node/zone spreading, zero-unavailable rolling updates, PDB, optional HPA, Restricted Pod Security, private management traffic, and bounded scratch/resources
- add Terraform roots for S3/EKS Pod Identity, GCS/GKE Workload Identity Federation, and Azure Blob/AKS Workload ID, with dedicated versioned storage and least-privilege data-plane identities
- reject static provider credentials, custom storage endpoints, unsigned/cleartext storage modes, direct LoadBalancer exposure, unrestricted NetworkPolicy peers, mutable image inputs, unsafe disruption budgets, and chart-owned metadata overrides
- keep ECS/Fargate as an explicit evaluation profile because its 120-second replacement limit cannot preserve Crab's ten-minute drain contract; exclude Lambda from the streaming data plane

### Production operation

- require OIDC outside loopback, stable shared state keys, storage-backed startup/readiness preflight, cluster-wide CAS transfer admission, graceful drain, request IDs, bounded Prometheus metrics, PodMonitor, alert rules, and an operations/restore runbook
- qualify local persistence, LFS range resume and locking, complete-root cold restore, and process loss during an in-flight native push through the real Compose/Caddy/RustFS stack
- add a provider-neutral live gate that verifies two-zone placement, provider-native workload-identity admission on original and replacement pods, management NetworkPolicy enforcement, OIDC ingress, direct traffic through multiple replicas, Git/LFS/locking consistency, and uninterrupted reads during a complete rollout
- issue an attested schema-3 receipt bound to provider, release source, image/chart digests, installed chart version, ServiceAccount, workload-identity mechanism, repository result, and pod replacement evidence

### Release supply chain

- publish exact-source AMD64/ARM64 images and the matching OCI Helm chart only from annotated `crab-http-server-vX.Y.Z` tags on `main`
- generate SBOM and provenance, reject fixable HIGH/CRITICAL runtime findings, attest image and chart, and refuse existing immutable tags or chart versions
- publish a non-latest GitHub Release containing the chart package plus a signed JSON deployment record that binds source, registry digests, platforms, and the package checksum
- make the protected live workflow verify that record and both registry artifacts and derive expected coordinates automatically; no operator-pasted digest
- require separate protected EKS, GKE, and AKS environments so cloud identities, reviewers, tag policy, and evidence remain isolated

## Verification

Current local evidence includes:

- exact-head `cargo test -p crab-http-server --locked`: 99 library tests and 4 binary tests passed; 3 environment-gated tests ignored; exact-head strict Clippy and workspace rustfmt passed
- exact-head repository frontend production build; real Docker Compose/Caddy/RustFS journeys for bootstrap, persistence, restart, restore, LFS resume/locking, and abrupt receive loss
- strict Helm lint and Kubernetes 1.29 kubeconform validation for EKS, GKE Standard, GKE Autopilot, and AKS
- Helm negative rendering tests for exposure, credentials, configuration, availability, networking, image, and metadata policy
- Terraform fmt/init/validate and native mocked-provider tests for AWS, GCP, and Azure storage/identity boundaries
- pinned ShellCheck and actionlint for deployment and release automation
- three valid and three mutation/rejection fixtures for EKS Pod Identity, GKE Workload Identity Federation, and AKS Workload ID admission
- positive and negative signed-release-record contract tests, including image, OCI chart, and downloadable chart-package digests
- final non-root multi-stage image inspection, target-native ARM64 build, pinned Trivy scan with no fixable HIGH/CRITICAL findings, and SPDX SBOM generation

Exact-head GitHub checks are currently queued. No baseline, snapshot, or expected-failure file was changed to silence a gate.

## Production qualification boundary

This PR provides the deployable implementation and repeatable evidence machinery. It does **not** claim that a provider is release-qualified yet. That requires:

1. merge to `main` and the first successful annotated server release;
2. separately protected EKS, GKE, and AKS qualification environments and cluster-scoped OIDC identities;
3. one successful attested live receipt per provider for the exact release;
4. provider version-selected restore and operational rollback drills before critical repositories are admitted.

The repository currently has no Crab live-qualification environment configured, so those external receipts cannot be produced from this PR.

## Review focus

- Is the catalog-backed discovery boundary the best replacement for per-repository configuration?
- Do the Terraform roles and admitted pod checks prove the intended provider-native identity path without static credentials or node fallback?
- Do the Helm rejection guards and live CNI probe fail closed under unsafe exposure or placement?
- Does the signed release record provide one unambiguous image/chart handoff without mutable resolution?
- Are remaining production claims stated narrowly enough relative to the evidence?

## AI assistance

OpenAI Codex assisted with implementation, documentation, and local verification.


forhappy:
Storage-contract hardening added in `7b547574`: both `serve` and `storage-probe` now fail before listener ownership unless the configured workload can read the catalog, perform a bounded list, claim and release a conditional CAS slot, write and delete a unique probe object, and observe post-delete absence. Interrupted delete probes are bounded by the existing one-day auth-prefix lifecycle. Local verification: 99 library tests passed (3 live-environment tests ignored), 4 CLI tests passed, strict Clippy passed, and rustfmt passed. Exact-head container/RustFS qualification: https://github.com/crabbuild/crab/actions/runs/34738328523.

forhappy:
Follow-up `b1b95e1f` makes the container gate list the RustFS preflight prefix after `storage-probe` and require zero live objects, so cleanup is now asserted rather than inferred. Actionlint 1.7.11 and Compose parsing pass locally. Superseding exact-head run: https://github.com/crabbuild/crab/actions/runs/34738478985.

forhappy:
Live provider evidence is now automatable. `005d7459` adds a protected manual workflow for EKS/GKE/AKS using short-lived GitHub OIDC, explicit rollout approval, private-cluster runner modes, exact deployed-image matching, receipt verification, and 90-day artifact retention. `c10d66a4` requires the workflow to run from an annotated server release tag reachable from `main`. `b80e4141` adds a packaged least-privilege Kubernetes RBAC example with no Secret or storage access and validates all four RBAC resources with Kubernetes 1.29 kubeconform. Local Actionlint, Bash parsing, ShellCheck, immutable-image rejection, and kubeconform checks pass. Exact-head container run: https://github.com/crabbuild/crab/actions/runs/34739224035.

forhappy:
Production deployment hardening update (`c6dfc3ec`):

- Removed `s3:ListBucketMultipartUploads`, which AWS does not support with the `s3:prefix` condition used by the role.
- Removed `s3:ListMultipartUploadParts`; `object_store` tracks part identifiers client-side and does not invoke either list operation for Crab uploads.
- Retained only prefix-restricted `ListBucket` and object-scoped get/put/delete/multipart-abort.
- Added a Terraform contract test that pins the exact runtime action sets.
- Documented the EKS/GKE/AKS runtime grants and boundaries.

Local proof using the CI-pinned Terraform 1.13.3 image: init, validate, and all 3 AWS plan tests pass. New exact-head CI: https://github.com/crabbuild/crab/actions/runs/34739721239

AWS permission contract: https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html

forhappy:
Additional production hardening at `f183e626` and `67991147`:

- GKE now grants `roles/storage.objectUser` at the dedicated bucket instead of `roles/storage.objectAdmin`. The narrower predefined role retains object read/list/create/update/delete and multipart operations but removes object-IAM administration.
- Added a mocked Terraform plan assertion that pins the GKE runtime role.
- Disabled persisted checkout credentials in the server container and release workflows; neither needs an authenticated Git write after checkout.

Local proof:
- GCP Terraform init and validate passed with the CI-pinned Terraform 1.13.3 image.
- All 6 GCP plan tests passed.
- Actionlint 1.7.11 passed all three server workflows.
- `git diff --check` passed.

Exact-head container run: https://github.com/crabbuild/crab/actions/runs/34739860196
Google role contract: https://cloud.google.com/storage/docs/access-control/iam-roles#storage.objectUser

forhappy:
Storage transport hardening at `d3cb9e86`:

- Added a deny-only S3 bucket policy for requests where `aws:SecureTransport` is false.
- Excluded AWS service principals from the network-context deny, matching AWS guidance for service-to-service calls whose transport context can be redacted.
- Covered both the bucket and all object resources.
- Added a Terraform contract test for both required conditions and documented the cross-provider transport boundary.

Local proof with the CI-pinned Terraform 1.13.3 image: init, validate, and all 4 AWS plan tests pass; `git diff --check` passes.

AWS guidance: https://docs.aws.amazon.com/AmazonS3/latest/userguide/amazon-s3-policy-keys.html

forhappy:
Container supply-chain hardening at `aaa9d086`:

- The root Docker context now excludes `.env` variants, common private-key formats, and SSH private-key names after all source-tree allow rules.
- Container CI asserts every deny pattern remains present.
- The reference documents the remote BuildKit cache boundary.

Direct Docker proof used synthetic `crates/crab-http-server/.env.production` and `crab/id_ed25519` sentinels: both were absent after `COPY .`, while the server Cargo manifest and repository frontend manifest remained present. The synthetic files were then removed. Actionlint 1.7.11 and `git diff --check` pass.

forhappy:
Ingress production contract at `e77c3efb`:

- The checked-in ingress-nginx team overlay now disables request and response buffering, uses HTTP/1.1 upstream, delegates body-size limits to Crab, applies 630-second upstream read/write idle timeouts, and requires HTTPS redirect.
- The Helm guide explains which guarantees must be reproduced for a different ingress controller.
- Container CI asserts every rendered annotation so the easy setup cannot silently regress to ingress-nginx's 1 MiB request limit, request buffering, or 60-second timeouts.

Local proof: strict Helm lint passed; all seven annotations rendered exactly; kubeconform validated 8/8 Kubernetes 1.29 resources; Actionlint 1.7.11 and `git diff --check` passed.

Ingress-nginx contract: https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/

forhappy:
Hardened the live Kubernetes evidence chain in `9ea174fa`:

- qualification receipts now use schema 2 and bind the deployed digest to the stable server release tag and source commit
- receipts enumerate the successful OIDC, cross-replica Git/LFS, durable-lock, zero-unavailable rollout, and post-rollout checks, including zero failed probes
- the protected EKS/GKE/AKS workflow verifies the official image's GitHub provenance against the exact tag, source commit, release workflow, and GitHub-hosted builder before touching the cluster
- the workflow signs the verified receipt and retains both it and an offline Sigstore bundle
- server artifact publication now has one canonical trigger: an annotated server release tag push
- CI covers invalid image, release-tag, and source-commit inputs; Bash, ShellCheck, Actionlint, and receipt-verifier probes pass locally

This makes receipts tamper-verifiable; it does not substitute for the still-required real provider runs.

forhappy:
Added two production-hardening slices after the setup walkthrough:

- `d3339291`: the PDB now uses `unhealthyPodEvictionPolicy: AlwaysAllow`, preserving `minAvailable` for healthy replicas without allowing a permanently unready pod to block node drains; the live gate verifies it.
- `8a3ab7f1`: the documented team setup pins the dedicated namespace to Restricted Pod Security for enforce/audit/warn at v1.29 or newer. Live qualification now verifies those labels, and its ClusterRole can only `get` the named namespace plus individual hosting nodes.

All three provider renders pass strict Helm lint and Kubernetes 1.29 kubeconform (8/8 each). The updated script passes Bash and ShellCheck; workflow changes pass Actionlint; the qualification RBAC passes kubeconform (4/4).

forhappy:
Production Helm hardening added in `c9145075`:

- rejects provider-prefixed cloud environment overrides except `AWS_REGION` / `AWS_DEFAULT_REGION`
- rejects generic object-store credential, credential-source, endpoint, cleartext, signature-bypass, and storage-authority aliases case-insensitively
- rejects duplicate environment names and constrains the `extraEnv` schema
- rechecks the admitted Deployment during live qualification and records `checks.workload_identity_only` in the signed receipt
- documents the workload-identity-only boundary in the Helm guide, deployment guide, and reference

Local proof:
- strict Helm lint passed for EKS, GKE, and AKS
- all three rendered profiles passed strict kubeconform against Kubernetes 1.29 (8/8 resources each)
- literal, lowercase, generic endpoint, and Secret-reference credential injection cases were rejected
- safe `AWS_REGION`, `AWS_DEFAULT_REGION`, `RUST_LOG`, and proxy-style settings remained accepted
- ShellCheck, Bash syntax, Actionlint, jq policy probes, and `git diff --check` passed

forhappy:
Live network-policy qualification added in `449cb76b`:

- the portable EKS/GKE/AKS gate now launches a short-lived Restricted, non-root, tokenless, read-only peer pod
- qualification fails if that ordinary peer can reach the server management listener on port 8789
- the probe image is digest-pinned and runs with dropped capabilities and bounded resources
- cleanup removes the probe on success and every exit path
- qualification RBAC adds namespace-scoped pod create/delete only; it still grants no Secret or object-storage access
- the signed receipt and verifier now require `checks.management_network_isolation`

Local proof:
- Bash syntax, ShellCheck, and Actionlint passed
- the inline probe manifest passed strict kubeconform against Kubernetes 1.29
- the updated four-resource qualification RBAC manifest passed strict kubeconform
- the pinned probe image contains curl and runs read-only as UID/GID 65534 with all capabilities dropped

Effective CNI enforcement still requires the protected live provider run; static validation cannot substitute for that receipt.

forhappy:
Provider Terraform security parity added in `d7dd4637`:

- Azure now uses AzureRM 5.5's supported `public_network_access` field instead of the deprecated boolean field scheduled for removal in v6
- Azure plan tests now enforce OAuth-only access, disabled shared keys, TLS 1.2+, infrastructure encryption, private Blob access, versioning, private-endpoint mode, one-day auth-only lifecycle scope, container-scoped data-plane RBAC, and exact AKS federation issuer/audience/subject
- AWS plan tests now enforce all S3 public-access blocks, bucket-owner enforcement, default encryption, versioning, and exact EKS Pod Identity placement
- GCP plan tests now enforce no force deletion, public-access prevention, uniform bucket-level access, versioning, and exact GKE workload-identity membership

Local proof with locked Terraform 1.13.3 and provider lockfiles:
- AWS: 6 passed
- GCP: 8 passed
- Azure: 5 passed
- all three modules passed `terraform validate`
- recursive `terraform fmt -check` and `git diff --check` passed

forhappy:
Provider Terraform regression coverage added in `d7dd4637`:

- Azure now uses AzureRM 5.5's supported `public_network_access` argument instead of the deprecated field removed in v6
- Azure tests enforce OAuth-only storage access, shared-key disablement, TLS 1.2+, infrastructure encryption, private container access, versioning, private-endpoint public-network disablement, auth-only expiration scope, container-scoped RBAC, and exact AKS federation identity
- AWS tests now enforce all public-access blocks, bucket-owner enforcement, default encryption, versioning, and exact EKS Pod Identity placement
- GCP tests now enforce public-access prevention, uniform bucket access, versioning, force-delete prevention, and exact GKE workload-identity membership

Locked Terraform proof:
- AWS: 6/6
- GCP: 8/8
- Azure: 5/5
- all modules validate; recursive format check passes

forhappy:
Added an operator-ready release handoff in `39f91f19`:

- every tagged server release now retains the packaged Helm chart, a JSON record binding tag/source/image digest/chart digest/platforms, and an offline attestation bundle
- the workflow verifies the signed manifest before publishing it
- server releases are explicitly not marked as the repository's main “Latest” release
- deployment and operations docs now use the signed record as the canonical source for immutable coordinates
- the team quick-start now separates one-time platform/identity work from per-release service-owner work and makes clear that repositories are catalog-discovered, not configured per replica

Local proof: actionlint v1.7.11 passes; the manifest generator/validator accepts a complete record and rejects malformed source/digest fields; `git diff --check` passes.

forhappy:
Follow-up hardening in `741d219e`: extracted release-record generation from the tag-only workflow into `deploy/generate-release-manifest.sh` and made ordinary container PR CI execute the same contract. CI now proves one complete record and rejects mismatched versions, malformed source commits, tagged image repositories, wrong chart repositories, and malformed digests before any release tag exists. The generator passes Bash syntax validation and pinned ShellCheck; both affected workflows pass actionlint v1.7.11.

forhappy:
Setup simplification in `f4fae911`: the protected EKS/GKE/AKS live workflow no longer accepts a manually pasted `expected_image`. It downloads and verifies the selected tag's signed release record, validates its exact tag/source/repository/digest/platform contract, derives the immutable image and chart references, verifies both registry attestations, and uses the derived image for deployed-state and receipt checks. Explicit mutation approval remains required. Updated the Kubernetes guide and operations runbook. Actionlint and a generated-record validation probe pass locally.

forhappy:
Provider identity evidence is materially stronger in `715af173`:

- EKS receipts require the admitted Pod Identity credential URI, authorization-token file, projected `pods.eks.amazonaws.com` token, and mount on every ready pod
- GKE receipts require the KSA→Google service-account annotation and reject key-file/service-account environment overrides
- AKS receipts require the managed-identity client annotation, opt-in pod label, federated-token projection/mount, and injected client/tenant/token variables
- the gate repeats these checks after every pod is replaced
- receipt schema 3 binds the signed chart digest, installed chart/app version, ServiceAccount, and exact identity mechanism
- qualification RBAC adds only `get` on ServiceAccounts; no Secret or storage access

Local proof: three valid provider fixtures and three mutation/rejection fixtures pass; Bash syntax and pinned ShellCheck pass; actionlint passes; strict Helm lint passes for EKS/GKE/AKS; qualification RBAC validates 4/4 against Kubernetes 1.29; the packaged chart contains all verifier assets.

forhappy:
Supply-chain handoff tightened in `2c0776f7`: the signed release record now includes the SHA-256 of the downloadable Helm `.tgz`, not only the OCI manifest digest. Operator docs download all three release assets, verify the record attestation, verify the package bytes, and install the exact local package, avoiding a mutable version resolution between verification and rollout. PR CI covers the new field and rejects a malformed package digest. Actionlint, pinned ShellCheck, positive generation/checksum proof, and the rejection probe pass locally.

forhappy:
Qualification environment boundary tightened in `b812d00f`: the workflow no longer defaults to one shared environment, and the runbook recommends separate protected EKS/GKE/AKS environments so cloud identities, reviewers, tag policy, and evidence ownership remain isolated. A read-only repository check currently shows no `crab-http-server-live` environment, variables, or secrets, so no live provider claim has been made from this PR.

